PR #22, opened 2020-04-30
restructure package to have umbrella package and core control module

MarcoEidinger:
1 hour ago I had a discussion with @sstadelman to have an umbrella package and introduce a module for core controls


MarcoEidinger:
@sstadelman I changed your proposed name "FioriCore" to "FioriSwiftUICore" to  make it more clear that in contains core UI controls implemented with SwiftUI

MarcoEidinger:
Please change it if you don't like it

MarcoEidinger:
Ideally, we also change the reference from SAPFiori that it uses product FioriCharts only

PR #31, opened 2020-06-01
fix unit test as fillColor has to be specified for ChartSeriesPalette

MarcoEidinger:


MarcoEidinger:
Hi @shengxu7 does this work for you?

PR #41, opened 2020-06-08
feat: various enhancement for column chart

shengxu7:


MarcoEidinger:
lgtm

one feedback: I saw you are sometimes using a merge commit instead of rebase. Is there a particular reason. I personally prefer to rebase (if I had simple changes) as I don't add an additional commit to the history.

PR #48, opened 2020-06-12
SegmentedControl rework

billzhou0223:
1. SegmentedControl dynamic height based on content size.
2. Few apis renaming.
3. Add `SegmentWidthMode.equal` mode.
4. Add `allowEmptySelection` api.
5. Bug fixes.


billzhou0223:
@MarcoEidinger I have some important changes for segmented control that I would like to merge. Didn't have chance to setup conventional commit yet, will do it next time.

We also need to make a new release (0.5.2) and I will update SAPFiori to consume it.

Issue #33, opened 2020-06-02
Support fetching data with Request object


sstadelman:
`DataFetcher` and `BaseBaseCard` should be configured to execute an asynchronous Request, which may load data from a local or remote resource.  When JSON data is fetched, it should be sent to the Request publisher.

The URL of the Request may be parameterized, using the `configuration.destinations` object of the `card`.  See [spec](https://openui5.hana.ondemand.com/test-resources/sap/ui/integration/demokit/cardExplorer/webapp/index.html#/learn/features/destinations).

The URL of the Request may also be parameterized using the `app.dataSources` object.  See [spec](https://openui5.hana.ondemand.com/test-resources/sap/ui/integration/demokit/cardExplorer/webapp/index.html#/learn/features/dataSources).

The URL of the request itself may be fully-qualified or relative.  According to the spec:

> If the URL is relative, it is going to be resolved based on the page instead of the manifest base path.

However, since in this case there is no page, we need to evaluate the long-term behavior.  In the meantime, we should resolve either to the location of `manifest.json` on the remote host, or, for local bundles, relative to the `manifest.json` data.

PR #56, opened 2020-06-17
feat: 🎸 Support fetching data with Request object

ShawnMa16:
DataFetcher and BaseBaseCard should be configured to execute an
asynchronous Request, which may load data from a local or remote
resource. When JSON data is fetched, it should be sent to the Request
publisher.

BREAKING CHANGE: 🧨 nothing

✅ Closes: #33

Do not merge yet


Sources/FioriIntegrationCards/Components/Common/Model/Manifest.swift
```diff
@@ -29,6 +32,35 @@ public class Manifest: Decodable, Identifiable, ObservableObject {
...
+    
+    public init(with fileName: String) throws {
+        let destinationDir = FileManager.default.temporaryDirectory.appendingPathComponent(fileName, isDirectory: true)
+        if FileManager.default.fileExists(atPath: destinationDir.path) {
+            baseURL = destinationDir.absoluteString
+        } else {
+            let sourceFile = Bundle.main.url(forResource: fileName, withExtension: ".zip")!
+            do {
+                try Zip.unzipFile(sourceFile, destination: FileManager.default.temporaryDirectory, overwrite: true, password: nil)
+                baseURL = destinationDir.absoluteString
```
sstadelman, line 44: I think that since this implementation is specific to the `.card.zip` scenario, we should update the init method to something like:

`public init(withCardBundleAt path: String) throws`

Otherwise, a developer might try to init with the path to `manifest.json`.

ShawnMa16: I see, that make sense

ShawnMa16:
closing this, will have another PR from my fork

Issue #60, opened 2020-06-19
project ‘FioriSwiftUI’ is damaged and cannot be opened due to a parse error


MarcoEidinger:
Using shell script `setupForCollaborators.sh` will cause corrupt file `FioriSwiftUI.xcodeproj`

> The project ‘FioriSwiftUI’ is damaged and cannot be opened due to a parse error. Examine the project file for invalid edits or unresolved source control conflicts.

MarcoEidinger:
occurs when using python 2.7 :(

tested with 3.7.6 and everything works fine :)

MarcoEidinger:
checking python module to manipulate xcode project file to add a run script for SwiftLint check execution

> this is a known issue of python 2.7, which is automatically resolved in python3

Source: https://github.com/kronenthaler/mod-pbxproj/issues/264#issuecomment-587139440

MarcoEidinger:
I’ll change the setupForCollaborators.sh in such a way that it attempts to add the run script when python version 3 is present. Users with python 2.7 have to manually add such run script.

FYI: CONTRIBUTING guidelines already point out to use 3.4+ and gives tip how to easily switch between python versions => https://github.com/pyenv/pyenv

PR #61, opened 2020-06-19
fix: project ‘FioriSwiftUI’ is damaged and cannot be opened due to a …

MarcoEidinger:
…parse error

occurs when using python 2.7 :(

change setupForCollaborators.sh in such a way that it attempts to
add the run script when python version 3 is present.

Users with python 2.7 have to manually add such run script.


MarcoEidinger:
Closes #60

Issue #69, opened 2020-06-26
Add FioriCharts Views to Xcode Library


MarcoEidinger:
In Xcode 12 it's possible to add custom views and modifiers to Xcode Library (https://developer.apple.com/videos/play/wwdc2020/10649) which improves API discoverability.

Currently, there is only one generic `ChartView` with a generic `ChartModel` to handle all kinds of chart types. Nevertheless, it is possible to define different entries for Xcode Libraries to serve code snippets for chart-specific  ChartView+ChartModel implementations

Example

```swift
#if swift(>=5.3)
@available(iOS 14.0, *)
struct LibraryContent : LibraryContentProvider {
    @LibraryContentBuilder
    var views: [LibraryItem] {
        LibraryItem(
            ChartView(ChartModel(chartType: .stock, data: [[16.52, 15.50, 18.67, 16.70, 16.90]], titlesForCategory: [["2020-01-06 10:00:00", "2020-01-06 10:30:00", "2020-01-06 11:00:00", "2020-01-06 11:30:00", "2020-01-06 12:00:00"]])),
            title: "Stock Chart"
        )
    }
}
#endif
```

From an API consumer perspective, we probably want to offer chart-type specific convenience Chart Views in the future. This is not the scope of this feature request. If so then the custom views added to Xcode Library can easily be updated.

MarcoEidinger:
Resolved by #70

MarcoEidinger:
Fixed #70

PR #70, opened 2020-06-26
feat: add custom views of FioriCharts to Xcode library

MarcoEidinger:
Relevant for Xcode 12 and beyond


MarcoEidinger:
Resolves #69

MarcoEidinger:
I only add one item per chart type into the Xcode library. Theoretically, we could add for any parameter variation of a chart type a new library item.  But I believe that's overkill. For new chart types (waterfall, etc) this code would need to be enhanced. If @shengxu7 has no objections I'd like to merge this PR

PR #91, opened 2020-08-01
refactor: being able to run Examples app w/o manual prerequisite step

### LINKED ISSUE #90: Being able to run Examples app without manual prerequisite steps
Running [examples](https://github.com/SAP/cloud-sdk-ios-fiori#examples) requires the manual prerequisite step of creating an Xcode project for a workspace. This is inconvenient for non-collaborators and hinders exploration.

The goal is to add the package as a local dependency to Xcode project which still allows contributors to easily test local changes

MarcoEidinger:
add the package as a local dependency to Xcode project which still allows contributors to easily
test local changes

fix #90


MarcoEidinger:
- With this PR you don't need to use a workspace to run the examples app
- Local changes to package do get automatically reflected in examples app
- For local standalone development, you can also open the swift package by double-clicking on the `Package.swift` file. The only thing you would miss (in contrast to using a generated Xcode project) would be SwiftLint checks
- For CI we'll keep the process of generating an Xcode project file as it has also other benefits

Issue #93, opened 2020-08-04
ChartDetailView error when running in iOS 14


alisdair73:
Hi All

Found some odd behaviour when running the Examples app under iOS 14 beta. There must be something different happening in Swift when it handles the onTapGesture closure - meaning that the currentModel is not being set (or not in the correct self?) when the ChartDetailView is called. I tried using [self] in for the closure but made no difference. Fixed it briefly using a global but this is just a dodgy work around :) 

Works perfectly under iOS 13 - so bit of an odd one.

iOS 14 error attached below

Thx!

![Screen Shot 2020-08-05 at 8 08 45 am](https://user-images.githubusercontent.com/503165/89350652-c3499d80-d6f3-11ea-8ff4-05ad89954999.png)

MarcoEidinger:
Thanks for reporting the issue! We haven't fully tested the chart library with iOS 14 beta so please excuse such issues. 

Yes, it's a weird issue and need further investigation to fully resolve it. We are currently busy adding more features for iOS 13+ so might take more time.

To avoid the crash I applied a dodgy workaround :) with #94

![weiredissue](https://user-images.githubusercontent.com/4176826/89357665-fcdbc380-d675-11ea-912d-717a6f4e0af7.gif)

In Xcode 12 beta you have to select a different example of the same chart type and then you can switch back and test the configuration of the example in question

As this is not a fix but a workaround we haven't created a new release. You can consume the master branch.

We'll keep the issue open until we have a better resolution. Hopefully, we can start on iOS 14 support soon.

alisdair73:
Hi Marco

Thanks again for the quick feedback/resolution - I'd been using the XCode 12 beta for Widgets and had seen the crash a couple of times - was only yesterday that I tested in XCode 11 (i'd see your demo on the SAP Code Talk so knew it worked) and found that that's what the difference was.

I'm very keen to see what the final resolution will be - my Swift is not so great and this has made me to realise that I don't fully get escaping closures (among other things)...

Thanks again.

MarcoEidinger:
PR #102 is fixing the observed behavior but the root cause seems a Xcode 12 beta bug. According to https://developer.apple.com/forums/thread/653247 this was already reported. Hopefully, this gets resolved so that the necessary code changes to circumvent the issue can be removed.

PR #94, opened 2020-08-05
test: fix crash in Examples app when selecting a chart

MarcoEidinger:
iOS 14 issue

root cause still under investigation

workaround is to tap different example of same chart type and
then it works


MarcoEidinger:
Relates #93

Issue #98, opened 2020-08-05
CollectionView in the Integration Cards was created multiple times.


ShawnMa16:
**Describe the bug**
CollectionView in the Integration Cards was created multiple times.

**To Reproduce**
Please build and run the example app with iPhone.
1. Go to ‘Integrations Cards’
2. Click on ‘list_dataInCardLevel’

**Expected behavior**
On iPad, you will see an endless print statements in the console.

**Screenshots**

**Mobile (please complete the following information):**
 - Device: iPad Pro 11
 - OS: iOS 14
 - Swift Package: Integration Cards
 - Version master branch

Issue #97, opened 2020-08-05
CollectionView in the Integration Cards is flashing


ShawnMa16:
**Describe the bug**
CollectionView in the Integration Cards is flashing.

**To Reproduce**
Please build and run the example app with iPhone.
1. Go to 'Integrations Cards'
2. Click on 'Collection View' tab

**Expected behavior**
You will see the collection view screen is flashing.

**Screenshots**

**Mobile (please complete the following information):**
 - Device: Phone 11
 - OS: iOS 14

PR #101, opened 2020-08-07
refactor: use LazyVGrid instead custom CollectionView in ExamplesApp

MarcoEidinger:
use LazyVGrid in iOS14 only and get rid of CollectionView in iOS13

re #97, re #98


MarcoEidinger:
Hi, I am getting rid of the custom CollectionView. As Stan told me it never worked properly. Let's use in iOS 14 a built-in SwiftUI component to illustrate cards in a grid/collection manner. In iOS 13 we don't show such visualization.

I am also working on fixes for dedicated card issues which can be observed in iOS 14 / Xcode12 beta

MarcoEidinger:
Works when using Xcode 11.6 to run on iOS 13.6
Works when using Xcode 12 to run on iOS 14
Crashes when using Xcode 12 to run iOS 13.6 :(

I am looking for a solution but you have an idea please let me know.

PR #103, opened 2020-08-07
refactor: wrap LazyVGrid in AnyView

MarcoEidinger:
to avoid crash when using Xcode 12 to run on iOS 13.x
see https://developer.apple.com/forums/thread/650818


MarcoEidinger:
relates #101

PR #104, opened 2020-08-07
add all integration card types in Examples app and fix some Xcode 12 / iOS 14 bugs

MarcoEidinger:
in addition fix issue in the header of integration card as status specified with {} placeholder shall be ignored (that's the behavior in web anyway)

### DIFF
--- a/Apps/Examples/Examples/FioriIntegrationCards/SampleData/InlineTestCases.swift
+++ b/Apps/Examples/Examples/FioriIntegrationCards/SampleData/InlineTestCases.swift
@@ -10,9 +10,11 @@ import Foundation
 import FioriIntegrationCards
 
 enum InlineTestCases: String, CaseIterable, Identifiable {
-    case analytical = "analytical_dataInCardLevel"
-    case list       = "list_dataInCardLevel"
-    case timeLine   = "timeline_dataInCardLevel"
+    case analytical = "analytical"
+    case list       = "list"
+    case timeLine   = "timeline"
+    case object     = "object"
+    case table      = "table"
     
     var id: String {
         return rawValue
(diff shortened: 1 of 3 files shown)


MarcoEidinger:
@sstadelman even I improved the table card (by not showing a cryptic status) it's far from perfect.

Screen Shot 2020-08-07 at 6 29 43 AM

Do you wanna create an issue to get closer to the web layout?

Screen Shot 2020-08-07 at 6 29 48 AM

Sources/FioriIntegrationCards/Components/Timeline/UI/TimelineItemView.swift
```diff
@@ -29,7 +29,6 @@ struct TimelineItemView: View {
             VStack {
                 SafeView(self.model.icon)
                     .frame(width: iconWidth, height: iconWidth, alignment: .top)
-                    .padding(.bottom, -30)
```
ShawnMa16, line 32: Removing this will break the existing UI layout for TimelineItemView.

MarcoEidinger: Can you test locally? For me it looked fine but if you have a better way to fix issue I am all ears :)

ShawnMa16:
I’ve tested locally, it looks different than before after removing the code.
Thanks!
Shawn m.

Sources/FioriIntegrationCards/Components/Timeline/UI/TimelineItemView.swift
ShawnMa16: I’ve tested locally, it looks different than before after removing the code.
Thanks

MarcoEidinger: Ok, I removed it. Then we have to find a different solution to solve the flickering for timelineCard

ShawnMa16: I tested it locally, the timeline card is not flickering.
Maybe the flickering is due to the collectionView

PR #120, opened 2020-08-26
feat: 🎸 change chartmodel selections api to support multiple selection in donut chart

shengxu7:


Sources/FioriCharts/Model/ChartModel.swift
```diff
@@ -311,13 +315,25 @@ public class ChartModel: ObservableObject, Identifiable, NSCopying {
      - single : Selects a single value in the currently selected series and category indices.
      - all : Selects one value in each series for the selected category index(es).
      */
-    @Published public var selectionMode: ChartSelectionMode = .single {
+    @Published public var _selectionMode: ChartSelectionMode = .single {
```
MarcoEidinger, line 318: Both properties `_selectionMode` and `selectionMode` are public. Can we expose only one to the consumer?

shengxu7: fixed.

PR #125, opened 2020-08-29
feat: 🎸 improve color palette

justinguo:


justinguo:
Please kindly find the PR for updated snapshot reference images at: https://github.com/SAP/cloud-sdk-ios-fiori-snapshot-references/pull/1

justinguo:
Strangely, `AnalyticalCardSnapshotTests` is passed locally without any changes.

image

MarcoEidinger:
> Strangely, `AnalyticalCardSnapshotTests` is passed locally without any changes.

That's the issue I mentioned via Slack. Currently, the color seems to alternate between 2 sets of colors. Reason unknown. I'll retrigger tests and let's cross fingers that it works. Anyway, I am able to merge even if this random failure cannot be solved

MarcoEidinger:
@justinguo  this commit contains breaking changes, correct? Because public API's are getting deleted (example: `func rgba(_ colorScheme...`)

MarcoEidinger:
@billzhou0223 ok for you?

@justinguo am I right with breaking changes? If so then I can squash the commit and add the necessary statement in new commit body

justinguo:
@MarcoEidinger Technically yes. But since all APIs in `HexColor` are used only internally for now and no other components consuming it so I guess it should be fine. So I did not put deprecated flags on the older APIs.

MarcoEidinger:
@justinguo I don't understand. `HexColor` is a public struct but will be used only internally by FioriSwiftUICore? So consumers using our package cannot really use it? Then why is struct public? I am not worried about the breaking change as we did not promise yet semantic versioning but I'd like to better understand

MarcoEidinger:
FYI: I merged it now without Bills review as Sheng was doing changes as well which invalidate snapshot reference images and I wanted that he can work with latest code matching latest reference images

billzhou0223:
> @justinguo I don't understand. `HexColor` is a public struct but will be used only internally by FioriSwiftUICore? So consumers using our package cannot really use it? Then why is struct public? I am not worried about the breaking change as we did not promise yet semantic versioning but I'd like to better understand

Ideally `HexColor` should be internal. But I am afraid it is used as return type of a method in public protocol so it has to be public as well. @justinguo Let's revisit this issue later. For now it is ok to merge this PR.

PR #132, opened 2020-09-06
chore: 🤖 color palette changes

billzhou0223:
No need to add these deprecated api in FioriSwiftUI. We can still keep them in SAPFiori using extension.


justinguo (approved):
Looks good

PR #137, opened 2020-09-18
ci: being able to publish release by triggering workflow manually

MarcoEidinger:


MarcoEidinger:
See https://github.com/MarcoEidinger/gh-workflow_dispatch-standard-version-recipe for more info

PR #157, opened 2020-10-20
fix: 🐛 improve hex color initialization to prevent crash on release build

justinguo:


justinguo:
I need to amend a commit message

PR #169, opened 2020-11-06
refactor: Moved strings files in FioriSwiftUICore to dedicated folder for organization

sstadelman:
Not sure of how these were be consumed, and if those paths would need to be updated.  But having all these language folders makes it difficult to work in the module.


MarcoEidinger:
Good idea! fyi to @sstadelman those string files are not yet used internally as we have not made the necessary changes to set minimum swift-tools-version to `5.3` but I plan to do this after 0.9 release as it is needed for other topics as well

MarcoEidinger (approved):

PR #175, opened 2020-11-10
chore: update README

MarcoEidinger:
- remove  "FioriCharts must support point and range selection, equivalent to the current implementation in SAPFiori"
- remove "This project is in development, and should not yet be used productively."

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,7 @@
 
 ***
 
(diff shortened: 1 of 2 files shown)


MarcoEidinger:
@sstadelman I'd like to merge this commit before releasing 0.9.0 so are those changes fine from your perspective?

PR #180, opened 2020-11-25
feat: integration cards: support list with numeric header

MarcoEidinger:
https://ui5.sap.com/test-resources/sap/ui/integration/demokit/cardExplorer/webapp/index.html#/explore/list/numeric

also fixes numeric header to consider state in addition
to trend to determine color

listnumeric


MarcoEidinger:
PR is ready for review and merge

Please note why failing checks can be ignored for now:
- code formatting check fails due to incorrect formatted FioriChart sources (not part of this PR). Sheng will fix these in a future PR
- snapshot testing due to
  - new test and missing reference image
  - changed behavior (consider state for KPI color determination and KPI unit of measurement shall be color-coded as well) 

 I'll record new snapshot reference images once this PR is merged (as I can do this via CI capabilities)

Sources/FioriIntegrationCards/Components/Common/Info.swift
```diff
@@ -0,0 +1,37 @@
+import SwiftUI
+
+public final class Info: Decodable, AnyBodyProducing {
+    var value: String?
+    var state: String?
+
+    func body() -> AnyView {
+        AnyView(
+            Text(self.value ?? "")
```
ShawnMa16, line 9: Probably we can use the `SafeText` here I guess?

ShawnMa16 (changes requested):

Sources/FioriIntegrationCards/Components/Common/Info.swift
MarcoEidinger: For some reason, it cannot find `SafeText`. Does it work for you?

ShawnMa16: I just tried, it won't show up either in my Xcode

ShawnMa16 (changes requested):

Sources/FioriIntegrationCards/Components/Common/Info.swift
MarcoEidinger: I pushed a commit which allows the use of `SafeText` without an instance of View and then I use it in this example

ShawnMa16 (approved):

PR #191, opened 2021-01-28
refactor: 💡 refactor XYAxisChart

shengxu7:


MarcoEidinger:
I'll fix the ci issue for snapshot tests and then retrigger the jobs to verify if snapshot tests are affected or not by your change

PR #194, opened 2021-02-01
feat: 🎸 support ShapeStyle for Chart series and categories

shengxu7:
Support ShapeStyle includes LinearGradient, AngularGradient,
RadialGradient, Color and ImagePaint.


MarcoEidinger:
@shengxu7 are all errors in failed snapshot tests expected (due to new feature) and is just ok to re-record snapshot images after merge (which I can trigger via CI)?

shengxu7:
> @shengxu7 are all errors in failed snapshot tests expected (due to new feature) and is just ok to re-record snapshot images after merge (which I can trigger via CI)?

@MarcoEidinger yes, pls re-record snapshot images. Due to the change in render engine, some are failed but the differences btw reference images and failed images are not distinguishable at all by human eyes.

Sources/FioriCharts/Model/ChartView.swift
```diff
@@ -104,6 +104,75 @@ public extension ChartView where Content == EmptyView {
...
+     .chartCategoryShapeStyle([0: [0: LinearGradient(gradient: Gradient(colors: [.red, .green, .blue]), startPoint: .bottom, endPoint: .top),
+                                   1: Color.yellow],
+                               1: [3: RadialGradient(gradient: Gradient(colors: [.red, .yellow, .green, .blue, .purple, .pink]), center: .center, startRadius: 0, endRadius: 50),
+                                   4: ImagePaint(image: Image("flower"))]])
+     */
+    func chartCategoryShapeStyle(_ value: [Int: [Int: Any]]) -> some View {
+        environment(\.chartCategoryShapeStyle, value)
+    }
+}
+
```
MarcoEidinger, line 175: I'd like to avoid `Any` as much as possible in public API. Are alll those types of`SwiftUI.ShapeStyle`? I understand that protocol 'ShapeStyle' can only be used as a generic constraint because it has Self or associated type requirements. But wouldn't be a a type erasure help here?

@billzhou0223 what's your point of view?

billzhou0223: @shengxu7 
Instead of using `Any` in method signature, the better way might be to enforce the type check at compilation time by using a type eraser like this:

```
public struct AnyShapeStyle {
    public let base: Any
    
    public init<S>(_ base: S) where S : ShapeStyle {
        self.base = base
    }
}

func chartSeriesShapeStyle(_ value: [Int: AnyShapeStyle]) -> some View
```

shengxu7: Changed it based on your suggestions.

billzhou0223 (approved):

PR #199, opened 2021-02-10
Latest source code generator enhancements 

MarcoEidinger:
- `post` phase to generate ViewModel compositions
- being able to generate standard component interfaces containing functions
- being able to generate non `@ViewBuilder` injectable APIs
- experimental examples in test app


MarcoEidinger:
@billzhou0223 as promised my changes. SwiftLint violations can be ignored imo as those are in experimental APIs only. I recommend to `squash and merge`. I will do this tomorrow before lunch time until I hear a veto from you.

PR #208, opened 2021-02-24
fix: 🐛 Selection is not preserved if not return to top level

billzhou0223:


Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
```diff
@@ -89,9 +89,29 @@ public struct ListPickerItemConfiguration {
...
+                                      @ViewBuilder rowContent: @escaping (Data.Element) -> RowContent)
+        where Data: RandomAccessCollection, RowContent: View, Data.Element: Identifiable, ID == Data.Element.ID
+    {
+        let id = \Data.Element.id
+        self.init(data, id: id, children: children, selection: selection, rowContent: rowContent)
+    }
 }
 
 public extension ListPickerItemConfiguration {
+    /// Creates a configuration object from a collection of `String` which supports both signle-level and multi-level picker with the ability to select multiple items.
```
MarcoEidinger, line 111: Spelling mistake `signle`

```suggestion
    /// Creates a configuration object from a collection of `String` which supports both single-level and multi-level picker with the ability to select multiple items.
```

MarcoEidinger (changes requested):

PR #210, opened 2021-03-03
chore: 🤖 support use of custom view-returning function builder in mod…

MarcoEidinger:
…el generation

### DIFF
--- a/sourcery/.lib/Sources/utils/Array+Variable.swift
+++ b/sourcery/.lib/Sources/utils/Array+Variable.swift
@@ -55,7 +55,13 @@ public extension Array where Element == Variable {
      ```
      */
     var viewBuilderInitParams: [String] {
-        map { "@ViewBuilder \($0.trimmedName): @escaping () -> \($0.trimmedName.capitalizingFirst())" }
+        map {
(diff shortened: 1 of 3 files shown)


MarcoEidinger:
@shengxu7 if you pull this PR and use it to build all phases then the ViewModel code should be generated in the way you expect it (equivalent to the manual changes you shared with me via Slack)

shengxu7 (approved):

sourcery/.lib/Sources/utils/Array+Variable.swift
```diff
@@ -55,7 +55,13 @@ public extension Array where Element == Variable {
      ```
      */
     var viewBuilderInitParams: [String] {
-        map { "@ViewBuilder \($0.trimmedName): @escaping () -> \($0.trimmedName.capitalizingFirst())" }
+        map {
+            if let cfb = $0.resolvedAnnotations("customFunctionBuilder").first {
+                return "@\(cfb) \($0.trimmedName): @escaping () -> \($0.trimmedName.capitalizingFirst())"
```
billzhou0223, line 60: Since we are calling view builder closures in the body of init method, we should be able to remove `@escape` keyword in the parameter list.

PR #215, opened 2021-03-15
fix: 🐛 Fix signature drawing pad issue

fredsap:


justinguo:
@billzhou0223 In this PR basically Fred saved the latest progress, including fixes for issues: 1) drawing now will stay within the bound, 2) after save is clicked the drawing pad displayed correctly in a scroll view. In general it looks good to me. Please double check if this can be merged.

@fredsap I double checked for the failed PR task on Xcode 11. It seems like we need to explicitly use `self.` in some closures. Could you make those changes? Thanks.

billzhou0223 (approved):

justinguo:
@MarcoEidinger Do you think the PR job named `CI (Xcode 11) / build (pull_request_target) ` runs properly?

MarcoEidinger:
Indeed the commit (9ad93feb95b63c78deda1a19b5190d5c64a6a37a) which used for checkout

Screen Shot 2021-03-15 at 4 50 07 PM

is the commit which started this PR  and not the latest one. I don't know why this occurs. Will need more time for analysis but busy with other stuff.

@billzhou0223 you have the ability as admin to merge the commit if it's critical so this CI issue is not a showstopper

billzhou0223:
> Indeed the commit ([9ad93fe](https://github.com/SAP/cloud-sdk-ios-fiori/commit/9ad93feb95b63c78deda1a19b5190d5c64a6a37a)) which used for checkout
> 
> Screen Shot 2021-03-15 at 4 50 07 PM
> 
> is the commit which started this PR and not the latest one. I don't know why this occurs. Will need more time for analysis but busy with other stuff.
> 
> @billzhou0223 you have the ability as admin to merge the commit if it's critical so this CI issue is not a showstopper

Yeah this doesn't seem right. The latest commit ought to be checked out should be 141c081. @fredsap Can you create a new PR for your latest fix to see if it runs correctly there? If not we can force merge this one after you confirm iOS 13 is running without issue.

fredsap:
making a new one to see if it fixes issues

PR #221, opened 2021-03-22
chore: 🤖 use customFunctionBuilder in +Init.generated.swift

MarcoEidinger:
using `// sourcery: customFunctionBuilder=<Name>` had no effect to
optional initializers in <Component>+Init.generated.swift file. This
change ensures that instead of @ViewBuilder the actual @<Name> custom
function builder is used.


MarcoEidinger:
@shengxu7 please verify this solves the issue you mentioned to me

shengxu7 (approved):

PR #226, opened 2021-03-29
Support both UIImage and Image types in Signature View 

fredsap:


justinguo:
@fredsap Could you check and fix the comments from hound? Also could you fix the conflicts with main branch?

billzhou0223 (commented):
I have a new proposal. Instead of having two closures to pass Image and UIImage which might be a little confusing. We can create a result struct to wrap them up. And pass this result to `onSave` handler. @justinguo Please let me know how you like this idea.

```
extension SignatureCaptureView {
  public struct Result {
    let image: Image
    let uiImage: UIImage
  }
}
```

justinguo:
> I have a new proposal. Instead of having two closures to pass Image and UIImage which might be a little confusing. We can create a result struct to wrap them up. And pass this result to `onSave` handler. @justinguo Please let me know how you like this idea.
> 
> ```
> extension SignatureCaptureView {
>   public struct Result {
>     let image: Image
>     let uiImage: UIImage
>   }
> }
> ```

This looks good to me. @fredsap Could you adopt this solution for now? Also please clean up unused code for conversion between `UIImage` and `Image` (if any).

Sources/FioriSwiftUICore/SignatureView/SignatureHelper.swift
```diff
@@ -19,6 +27,19 @@ extension UIView {
...
+            bezierPath.move(to: point)
+        } else {
+            bezierPath.addLine(to: point)
+        }
+    }
+    return bezierPath
+}
+
 extension View {
     func snapshot() -> UIImage {
```
justinguo, line 44: @fredsap Are we using this extension method currently? If not we can remove it for now.

Sources/FioriSwiftUICore/SignatureView/SignatureHelper.swift
```diff
@@ -19,6 +27,19 @@ extension UIView {
     }
```
justinguo, line 27: @fredsap Same here.

justinguo:
@billzhou0223 In general this looks fine to me as of now. Do you think this is fine?

Sources/FioriSwiftUICore/SignatureView/SignatureHelper.swift
```diff
@@ -2,6 +2,14 @@ import Foundation
 import SwiftUI
 import UIKit
 
+/// Result holds an `Image` and a `UIImage` type for the same signature image
+public struct Result {
```
billzhou0223, line 6: I suggest making it a nested type of SignatureCaptureView. Otherwise you need to rename it to be more specific instead of calling it Result.

```
extension SignatureCaptureView {
  public struct Result {}
}
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/SignatureView/SignatureHelper.swift
fredsap: cant because drawing pad needs result

billzhou0223: What do you mean? It's just a type under SignatureCaptureView `SignatureCaptureView.Result`

justinguo (approved):
Looks good now. @billzhou0223 Do you think this looks fine now?

billzhou0223 (approved):

PR #227, opened 2021-03-30
feat: 🎸 ObjectItem

shengxu7:


Package.swift
```diff
@@ -22,6 +22,11 @@ let package = Package(
             name: "FioriIntegrationCards",
             type: .dynamic,
             targets: ["FioriIntegrationCards"]
+        ),
+        .library(
```
billzhou0223, line 26: Let's revert this change.

billzhou0223 (changes requested):
I recommend adding doc for public apis next time before creating PR.

Package.swift
shengxu7: Reverted.

billzhou0223 (approved):

MarcoEidinger (approved):

PR #231, opened 2021-04-05
fix: 🐛 Fix border and multi-gesture issues in Signature View 

fredsap:


billzhou0223 (changes requested):
Instead of hard coding the value, can we create a property in the drawing pad for the margin and use it in calculation. This makes it easier if we need to adjust the margin in the future. 

```
struct DrawingPad {
    let padding: EdgeInset
}
```

Sources/FioriSwiftUICore/SignatureView/DrawingPad.swift
```diff
@@ -40,6 +40,7 @@ struct DrawingPad: View {
     var strokeColor: Color
     var lineWidth: CGFloat
     var backgroundColor: Color
+    let signaturePadding: CGFloat = 20
```
billzhou0223, line 43: Any reason you use `CGFloat` here which only provides one single value for padding on four edges.

What if we want the padding to be different on different edges? That's why I suggest using `EdgeInsets` in last comment.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/SignatureView/DrawingPad.swift
fredsap: I see your point, but there's really isn't any way to make the padding different on different edges. That will mean different and specific calculations for the signature origin.

billzhou0223 (approved):

PR #241, opened 2021-04-13
refactor: 💡 change package structure

justinguo:


Apps/Examples/Examples/FioriSwiftUICore/ObjectItem/SingleActionFollowButton.swift
```diff
@@ -1,3 +1,4 @@
+import FioriSwiftUI
```
billzhou0223, line 1: Is this needed?

Sources/FioriCharts/Charts/Common/MicroChartDataItem.swift
```diff
@@ -1,4 +1,4 @@
-import FioriSwiftUICore
+import FioriThemeManager
```
billzhou0223, line 1: Actually we don't even need to import `FioriThemeManager` because the color api is defined in an extension of `Color`. Can you try if project builds successfully if we remove this?

billzhou0223 (changes requested):

justinguo:
@billzhou0223 I checked and removed all unnecessary imports. All others are required for now.

billzhou0223 (approved):

PR #258, opened 2021-04-30
feat: migrate sourcery infrastructure and generated apis.

billzhou0223:


billzhou0223:
@MarcoEidinger Please double check if the changes in github/workflow are expected.

Currently CI build is failing because of wrong swift tools version.

MarcoEidinger:
> @MarcoEidinger Please double check if the changes in github/workflow are expected.
> 
> Currently CI build is failing because of wrong swift tools version.

Hi @billzhou0223 , I don't believe you have copied all changes from github/workflow. `ci_xcode11.yml` requires an update as I did in https://github.com/SAP/cloud-sdk-ios-fiori/blob/migration/.github/workflows/ci_xcode11.yml. I kept the same name which is misleading (I am guilty hear) because we require Xcode 12 as our package requires swift-tools-version 5.3 (FYI: this has nothing to do with the fact to build for iOS 13)

Also, we should not copy the README of migration branch 1:1 but rather move the details about developing controls with source code generation into a separate markdown file

billzhou0223:
> > @MarcoEidinger Please double check if the changes in github/workflow are expected.
> > Currently CI build is failing because of wrong swift tools version.
> 
> Hi @billzhou0223 , I don't believe you have copied all changes from github/workflow. `ci_xcode11.yml` requires an update as I did in https://github.com/SAP/cloud-sdk-ios-fiori/blob/migration/.github/workflows/ci_xcode11.yml. I kept the same name which is misleading (I am guilty here) because we require Xcode 12 as our package requires swift-tools-version 5.3 (FYI: this has nothing to do with the fact to build for iOS 13)
> 
> Also, we should not copy the README of migration branch 1:1 but rather move the details about developing controls with source code generation into a separate markdown file

Made some changes based on your advice. Please take another review.

billzhou0223:
Merge this PR to unblock development from SAPFiori side. Furthur review is pending. @MarcoEidinger

PR #259, opened 2021-05-06
Add background color

ShawnMa16:


Sources/FioriSwiftUICore/DataTable/DataTable.swift
```diff
@@ -41,5 +42,11 @@ public struct DataTable: View {
         }
         .environmentObject(layoutManager)
         .environmentObject(dataManager)
+        .environment(\.backgroundColor, self.backgroundColor)
```
billzhou0223, line 45: No need to set environment here casue it's already set in modifier.

Sources/FioriSwiftUICore/DataTable/ItemView.swift
```diff
@@ -70,15 +70,13 @@ struct ItemView: View {
...
-//                .frame(width: (dataItem.size.width + contentInset * 2) * self.layoutManager.scaleX, height: dataItem.rowHeight * self.layoutManager.scaleY, alignment: .center)
-//                .background(backgroundColor)
                 
                 if index == (0, 0) && self.layoutManager.model.isFirstColumnSticky {
                     let offsetX: CGFloat = self.dataItem.size.width / 2 + contentInset
                     verticalDivider(offsetX: offsetX)
                 }
             }
-            .background(Color.white.frame(width: (self.dataItem.size.width + contentInset * 2) * self.layoutManager.scaleX, height: self.dataItem.rowHeight * self.layoutManager.scaleY))
+            .background(self.backgroundColor.frame(width: (self.dataItem.size.width + contentInset * 2) * self.layoutManager.scaleX, height: self.dataItem.rowHeight * self.layoutManager.scaleY))
```
billzhou0223, line 79: It's not very common to set frame for background. Maybe you should add the background to the item view instead of ZStack.

billzhou0223 (changes requested):

billzhou0223 (approved):

PR #260, opened 2021-05-07
SignatureCaptureView: set drawing area idealHeight and minHeight to 256

janhuachu:
This is needed for FUIInlineSignatureFormCell size requirement.


billzhou0223:
@janhuachu You can run `setupCollaborator.sh` which should install the commit hook for you. When you commit your changes in the future, run `git commit` and it will prompt the required fields to fill in so that your commit message complies with conventional commit protocol.

Sources/FioriSwiftUICore/SignatureView/SignatureCaptureView.swift
```diff
@@ -72,6 +72,8 @@ public struct SignatureCaptureView: View {
                         Color.preferredColor(.quarternaryFill).cornerRadius(10)
                         Text(NSLocalizedString("Tap to Sign", comment: "Tap to Sign")).foregroundColor(Color.preferredColor(.tintColor)).font(.body)
                     }
+                    .frame(idealHeight: 256)
+                    .frame(minHeight: 256)
```
billzhou0223, line 76: You can put these two parameters into the same frame modifier. 

```
.frame(minHeight: 256, idealHeight: 256)
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/SignatureView/SignatureCaptureView.swift
janhuachu: Thanks. After some more tests it sometimes still not work in the form cell.

PR #261, opened 2021-05-10
fix: SignatureCaptureView height and localization

janhuachu:


janhuachu:
After discussing with Bill, this PR will be closed without merge.
Will use modifier instead of init parameters.

PR #272, opened 2021-05-21
Add theming apis for Data Table

ShawnMa16:


Sources/FioriSwiftUICore/DataTable/DataTextItem.swift
```diff
@@ -9,9 +9,11 @@ public struct DataTextItem: DataItem, CheckBinding {
     /// String for text item.
     public var text: String
     /// Font for item.
-    public var font: Font
+    public var font: Font?
     /// Line limit for item.
     public var lineLimit: Int?
+    /// Foreground color for text item.
+    public var fontColor: Color?
```
billzhou0223, line 16: Usually `textColor` could be a better name.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/DataTable/DataTextItem.swift
billzhou0223: @ShawnMa16 Let's get this updated so that we can close this PR.

justinguo:
@ShawnMa16 Can you check if the issue in the failed PR jobs can be resolved? Thanks.

billzhou0223 (approved):

PR #282, opened 2021-06-02
refactor: 💡 add UIKit compatible side bar initializer

justinguo:


Sources/FioriSwiftUICore/Views/SideBar+View.swift
```diff
@@ -278,3 +302,43 @@ struct ListItemBackgroundSelectionStyle: ViewModifier {
         }
     }
 }
+
+/// :nodoc:
+/// Internal used by SDK for UIKit functionality only. NOT intended to used by developer.
+public class ListSelectedItem<Data>: ObservableObject where Data: RandomAccessCollection, Data.Element: Identifiable & Hashable {
```
billzhou0223, line 308: Can you add a underscore before it name to distinguish it from other public apis?

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/SideBar+View.swift
justinguo: 👍 Done

billzhou0223 (approved):

PR #298, opened 2021-06-15
feat: 🎸 support icon for status and substatus protocols

shengxu7:


Sources/FioriSwiftUICore/Views/TextOrIconView.swift
```diff
@@ -0,0 +1,34 @@
+import SwiftUI
+
+/// TextOrIconView to display TextOrIcon
+public struct TextOrIconView: View {
```
billzhou0223, line 4: `Text` can be initialized with both string and image so introducing a new wrapper view might not be neccessary.

billzhou0223 (changes requested):

billzhou0223 (approved):

PR #313, opened 2021-07-29
fix: 🐛 SignatureCaptureView xmark and line matching spec

janhuachu:
1. The current implementation has no xmark and signature line in the default state while the spec has.
2. The current implementation the xmark and signature line are center aligned in the middle while in the spec they are bottom aligned.
3. The current implementation the xmark and signature line color are not matching the spec in the active state.


billzhou0223 (approved):

billzhou0223:
@janhuachu Please update commit head in rel-6.1 in SAPFiori correspondingly :)

PR #318, opened 2021-08-18
refactor: 💡 fix downstream projects build issue

justinguo:


Sources/FioriSwiftUICore/DataTypes/FioriUIControlState+DataType.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Data types for `FioriUIControlState`. Equivalent to `UIControl.State`.
+public enum FioriUIControlState {
```
billzhou0223, line 4: Can we name it `ControlState` since normally we don't add fiori prefix in SwiftUI project?

We should also make it conforms to `OptionSet` just to make sure it works in the same way as `UIControlState`.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/DataTypes/FioriUIControlState+DataType.swift
justinguo: Sounds good.

justinguo: We cannot use `ControlState` as the name since there is a conflict. I use `ControlStateSet` instead since it is an `OptionSet`.

Sources/FioriSwiftUICore/DataTypes/FioriUIControlState+DataType.swift
```diff
@@ -1,15 +1,27 @@
 import Foundation
 
 /// Data types for `FioriUIControlState`. Equivalent to `UIControl.State`.
-public enum FioriUIControlState {
+public struct ControlStateSet: OptionSet {
```
billzhou0223, line 4: Name of `ControlStateSet` sounds weird to me. 

I found `ControlState` is already defined under `FioriSwiftUICore/DimensionSelector/ControlState`. Can you update it to adopt `OptionSet` and move the definition into the utility folder.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/DataTypes/FioriUIControlState+DataType.swift
justinguo: Cool. I changed that implementation and tested on simulator. Functionality is the same as before.

billzhou0223 (approved):

PR #323, opened 2021-09-21
fix: 🐛 Xcode 13 compilation error in TimelineItemView

MarcoEidinger:
fixes compilation error when using Xcode 13 due to 'ambiguous use of
.center'


MarcoEidinger:
@billzhou0223 once you approve I'd downport this fix and create 1.0.1 release including this fix only. The sidebar fix from Justin does not have to be downported as 1.0.0 does not contain that code.

PR #329, opened 2021-09-27
fix: 🐛 support chart for iOS Widget

shengxu7:
the root cause is iOS Widget view has to be pure swiftui view. but the chart uses a few ui gesture recognizers from UIKit since gestures from swiftui are not powerful enough. Because iOS Widget doesn’t need to support user interaction so it is fine to turn off those part of code.

### DIFF
--- a/Sources/FioriCharts/Charts/Common/GridLinesAndChartView.swift
+++ b/Sources/FioriCharts/Charts/Common/GridLinesAndChartView.swift
@@ -135,37 +135,39 @@ struct GridLinesAndChartView<Content: View, Indicator: View>: View {
             
             indicatorView
             
-            Background(tappedCallback: { point, chartRect, _ in
-                if !self.model.selectionEnabled {
-                    return
-                }
-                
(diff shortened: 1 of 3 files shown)


MarcoEidinger (approved):
Verified with https://github.com/MarcoEidinger/FioriChartsInWidgets using Swift Package from https://github.com/shengxu7/cloud-sdk-ios-fiori and its PR branch`widget`

Screen Shot 2021-09-27 at 12 54 31 PM

MarcoEidinger:
@billzhou0223 merge, cherry-pick to rel-1.0 and publish release `1.0.2` ? I can do it or if you would like (i.e. trying out the release workflow), just let me know.

billzhou0223 (approved):

billzhou0223:
> @billzhou0223 merge, cherry-pick to rel-1.0 and publish release `1.0.2` ? I can do it or if you would like (i.e. trying out the release workflow), just let me know.

I asked Sheng to update the API doc to warn developers not to turn on `userInteraction` for charts if they are used in widget extension. Otherwise the chart will not display. 

After it's checked in. We can do the cherry pick and publish a new release. Please go ahead on your side :)

MarcoEidinger (approved):

PR #380, opened 2021-11-24
fix: 🐛 dimension selector styling

justinguo:


Sources/FioriSwiftUICore/DimensionSelector/DimensionSelector.swift
```diff
@@ -257,16 +256,22 @@ extension DimensionSelector {
...
-                .foregroundColor(self.isEnable ? (self.isSelected ? self.segmentAttributes[.selected]?.textColor : self.segmentAttributes[.normal]?.textColor) : (self.segmentAttributes[.disabled]?.textColor))
+                .font(self.isEnable ? (isSelected ? segmentAttributes[.selected]?.font : self.segmentAttributes[.normal]?.font) : self.segmentAttributes[.disabled]?.font)
+                .foregroundColor(isEnable ? (isSelected ? self.segmentAttributes[.selected]?.textColor : self.segmentAttributes[.normal]?.textColor) : (self.segmentAttributes[.disabled]?.textColor))
+                .background(RoundedRectangle(cornerRadius: cornerRadius, style: .continuous)
+                    .strokeBorder(isEnable ? (isSelected ? self.segmentAttributes[.selected]!.borderColor! : self.segmentAttributes[.normal]!.borderColor!) : (self.segmentAttributes[.disabled]!.borderColor!), lineWidth: isSelected ? 1.0 : 0.33)
+                    .background(RoundedRectangle(cornerRadius: cornerRadius, style: .continuous).fill(backgroundColor)))
```
billzhou0223, line 274: Why there are two background modifiers? I think the sequence of the modifers should be as follows:

```
Text("Hello World")
                .background(Color.green)
                .border(Color.red)
                .cornerRadius(8, antialiased: true)
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/DimensionSelector/DimensionSelector.swift
justinguo: We need to have two background modifiers. One is for strokeBorder, the other is for fillColor. Both need to use `RoundedRectangle` with `cornerRadius`. Your implementation will not work: the cornerRadius will not take any effect in latest SwiftUI, the `background` color will fill a rectangle and not a rounded corner rectangle.

billzhou0223 (approved):

PR #386, opened 2021-12-13
fix: 🐛 use FioriSwiftUI package with version-based requirement

MarcoEidinger:
It was not possible to specify dependency for `FioriSwiftUI` package
with `upToNextMajor(from:)` or `upToNextMinor(from:)` due to the use of
commit-based requirement for Zip dependency. Switching to a stable fork
in Package.swift will fix that.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -21,11 +21,11 @@
       },
       {
         "package": "Zip",
-        "repositoryURL": "https://github.com/maparoni/Zip.git",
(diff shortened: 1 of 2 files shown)


MarcoEidinger:
Once merged I'll use `Release` GitHub action to create 2.0.1 on `rel-2.0` branch

(and cherry-pick to main)

billzhou0223 (approved):

PR #387, opened 2021-12-13
chore: Merge rel-2.0 into main.

billzhou0223:


MarcoEidinger:
Do you want to delete `UserConsentForm` implementation in main?

billzhou0223:
> Do you want to delete `UserConsentForm` implementation in main?

No, they should be kept in main. I reverted the change.

MarcoEidinger (approved):
the GitHub workflow to verify package on iOS 13 is no longer needed as we bump up minimum to iOS 14. I will remove this workflow soon.

PR #398, opened 2022-01-19
Make targets static

iamhansi:


iamhansi:
Oops, meant to do this on a fork. Closed

PR #409, opened 2022-03-17
fix: 🐛 Explicitly set Font.fiori() to use system font when registerFont() is not called.

billzhou0223:


MarcoEidinger (approved):
I tested this fallback handling and it works and it will solve the issue that previously the following statement did **NOT** produce a bold font **IF** `Font.registerFioriFonts()` was **NOT** called

```Swift
.font(Font.fiori(forTextStyle: .headline)).bold() // was not bold :(
```

MarcoEidinger:
@billzhou0223 thanks! I would need this published as `2.0.2` release to leverage the fix in `SAPCAI`

PR #417, opened 2022-04-18
fix: 🐛 [SN:CS20220002723901]Criticial orange label color

xiaoqinggrace:


billzhou0223:
Please update commit head in submodule in SAPFiori correspondingly.

PR #418, opened 2022-04-20
fix: [HCPSDKFIORIUIKIT-2043] resolve UIFont instance from Swift pass to SwiftUI can't be resized and empty text disable accessibility.

AllenCao001:


Sources/FioriSwiftUICore/DataTable/LayoutData.swift
```diff
@@ -143,7 +143,13 @@ class LayoutData {
                 let title = item.text
                 var uifont: UIFont
                 if let tmpUIFont = item.uifont {
-                    uifont = tmpUIFont
+                    // `item.uifont` is passed by developer, although is a preferred font but can't resize according to system in SwiftUI. So need to redefine the UIFont instance.
+                    if let styleName = tmpUIFont.fontDescriptor.fontAttributes[.textStyle] as? String {
```
billzhou0223, line 147: In order for the text to be resizable, developes are responsible for passing in a scaled font. We, as SDK, are not supposed to force this conversion.

billzhou0223 (changes requested):
@shengxu7

shengxu7 (approved):

Sources/FioriSwiftUICore/DataTable/LayoutData.swift
AllenCao001: At now, the issue is developers had init a `UIFont` instance via `UIFont.preferredFont()` in SDK and passed it to SwiftUI, but the text doesn't be resizable in SwiftUI. In order to fix the issue, so need to redefine `UIFont` in SwiftUI, it is as the above code.

billzhou0223 (approved):

AllenCao001:
When it can be merged into the main branch? Whether this pull request has any question? Because I need to fix the issue-2043.

PR #471, opened 2022-10-18
feat: 🎸 HCPSDKFIORIUIKIT-2219 a searchable list picker

xiaoyu0722:
A searchable list picker that support tracking alive or not.

### DIFF
--- a/Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
+++ b/Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
@@ -63,6 +63,26 @@ public extension ListPickerItem {
 public struct ListPickerItemConfiguration {
     let destinationView: AnyView
     
+    /// Creates a searchable configuration object from a collection of data which supports signle-level picker with the ability to select.
+    /// - Parameters:
+    ///   - data: The data for constructing the list picker.
+    ///   - id: The key path to the data model's unique identifier.
+    ///   - children: The key path to the optional property of a data element whose value indicates the children of that element.
+    ///   - selection: A binding to a set which stores the selected items.
+    ///   - searchFilter: The closure to filter the `data` in searching process. Request a boolen by the element and the filter key.
+    ///   - rowContent: The view builder which returns the content of each row in the list picker.
+    @available(iOS 15.0, macOS 12.0, *)
+    public init<Data, ID, RowContent>(_ data: Data,
+                                      id: KeyPath<Data.Element, ID>,
+                                      children: KeyPath<Data.Element, Data?>?,
+                                      selection: Binding<Set<ID>>?,
+                                      searchFilter: @escaping (Data.Element, String) -> Bool,
+                                      @ViewBuilder rowContent: @escaping (Data.Element) -> RowContent)
+        where Data: RandomAccessCollection, RowContent: View, ID: Hashable
+    {
+        self.init(data, id: id, children: children, selection: selection, isTopLevel: true, searchFilter: searchFilter, rowContent: rowContent)
+    }
+    
     /// Creates a configuration object from a collection of data which supports both signle-level and multi-level picker with the ability to select multiple items.
     /// - Parameters:
     ///   - data: The data for constructing the list picker.
@@ -124,6 +144,27 @@ public extension ListPickerItemConfiguration {
     }
 }
 
+private extension ListPickerItemConfiguration {
+    @available(iOS 15.0, macOS 12.0, *)
+    init<Data, ID, RowContent>(_ data: Data,
+                               id: KeyPath<Data.Element, ID>,
+                               children: KeyPath<Data.Element, Data?>?,
+                               selection: Binding<Set<ID>>?,
+                               isTopLevel: Bool,
+                               searchFilter: @escaping (Data.Element, String) -> Bool,
+                               @ViewBuilder rowContent: @escaping (Data.Element) -> RowContent)
+        where Data: RandomAccessCollection, RowContent: View, ID: Hashable
+    {
+        self.destinationView = SearchListView(data: data,
+                                              id: id,
+                                              children: children,
+                                              selection: selection,
+                                              isTopLevel: isTopLevel,
+                                              searchFilter: searchFilter,
+                                              rowContent: rowContent).typeErased
+    }
+}
+
 extension ListPickerItem {
     struct Row<ID: Hashable>: View where Value == EmptyView {
         private let content: Key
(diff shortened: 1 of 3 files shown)


Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
```diff
@@ -160,3 +183,70 @@ extension ListPickerItem {
...
+    let data: Data
+    let id: KeyPath<Data.Element, ID>
+    let selection: Binding<Set<ID>>?
+    let isTrackingLive: Bool
+    let rowContent: (Data.Element) -> RowContent
+
+    init(data: Data,
+         id: KeyPath<Data.Element, ID>,
+         selection: Binding<Set<ID>>?,
+         isTrackingLive: Bool,
```
billzhou0223, line 206: Rename to `isTrackingLiveChanges`

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
```diff
@@ -160,3 +183,70 @@ extension ListPickerItem {
         }
     }
 }
+
+@available(iOS 15.0, *)
+struct ListPickerSearchableView<Data: RandomAccessCollection,
```
billzhou0223, line 188: 1. Does your component support multi-level selection?
2. How can developers provide their own search criteria? (closure based)

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
xiaoyu0722: Currently, spec does not need multi-level selection, and I will double confirm with spec owner.
And for search criteria, I added a new `keypath` for filter data.

billzhou0223: The current implementation of ListPickerItem already supports multi-level selection. We can add search on top of it instead of aborting this existing feature.

Search criteria is not just a keypath, it may involve control flow and some logic. We should let developers pass list picker a callback so that we can use it to filter the dataset. The usage code could be something like:

```
ListPickerItem(data) { searchString, item in
    return item.name.contains(searchString)
}
```

xiaoyu0722:
```swift
public init<Data, ID, RowContent>(_ data: Data,
                                  id: KeyPath<Data.Element, ID>,
                                  children: KeyPath<Data.Element, Data?>?,
                                  selection: Binding<Set<ID>>?,
                                  isTrackingLiveChanges: Bool = false,
                                  criteria: @escaping (Data.Element, String) -> Bool,
                                  @ViewBuilder rowContent: @escaping (Data.Element) -> RowContent)
```
A criteria closure is required for searchable `ListPickerItemConfiguration`
Also an optional children is also provided, but currently, the search bar is separately for each level list.
The sublist will tracking changes, that developer only can confirm/cancel the changes at root list.
Spec is not ready for multi-level support, double confirmed with the spec owner(only single-level is required). Multi-level support may not be the final version.

billzhou0223: As a SDK component it ought to be extendable and we need to plan for requirement changes. Multi-level picker is very useful and we had customers ask for it back in the days. So it's nice if we provide it as an option.

Regarding to cancel/confirm action, that's a little tricky because we need to somehow store the selections temporarily until users confirm the selection or discard the changes if users cancel it at the top level. I feel this might require changes in the current architecture of list picker. I need to do some experiments on my side. In the meanwhile you can also think about it.

billzhou0223: I did some experiments. The search and confirm/cancel now works with multi-level selection. You may want to take a look and start from there.

xiaoyu0722: Emm, I see your branch.
Current implementation use `isTrackingAlive` instead of `isTopLevel` for the condition to show cancel/confirm actions.
```swift
if isTrackingAlive == true {
  no cancel/confirm actions
} else {
  show cancel/confirm actions
  use selections buffer
  and force tracking buffer alive.(no actions for the sub list actions)
  update selections by buffer once confirm tapped.
}
```
So same implementation?

billzhou0223: I don't think `isTrackingLiveChanges` should be tied to confirm/cancel logic. The selection binding property should only be updated when confirm is triggered. 

If you want to make `isTrackingLiveChanges` useful, you can add a callback or publisher which notifies the receiver whenever the selection changes. Do not add the interim changes into selection binding property since it is only for storing the final result.

billzhou0223: `isTopLevel` is for internal use only. It's not visible to developers.

xiaoyu0722: Got it, I remove the `isTrackingChangesAlive` first. I'm not sure if this will effect the actions.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
```diff
@@ -152,6 +144,27 @@ public extension ListPickerItemConfiguration {
...
+    init<Data, ID, RowContent>(_ data: Data,
+                               id: KeyPath<Data.Element, ID>,
+                               children: KeyPath<Data.Element, Data?>?,
+                               selection: Binding<Set<ID>>?,
+                               isTopLevel: Bool,
+                               searchFilter: @escaping (Data.Element, String) -> Bool,
+                               @ViewBuilder rowContent: @escaping (Data.Element) -> RowContent)
+        where Data: RandomAccessCollection, RowContent: View, ID: Hashable
+    {
+        self.destinationView = ListPickerSearchableView(data: data,
```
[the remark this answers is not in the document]
xiaoyu0722: fixed.

billzhou0223 (approved):

xiaoyu0722:
`Testing failed:
	Value of type 'List<Never, ForEach<[Data.Element], ID, _ConditionalContent<ListPickerItem<RowContent, EmptyView>, ListPickerItem<RowContent, EmptyView>.Row<ID>>>>' has no member 'searchable'
`
`searchable` is not imported in Xcode 12. Is this job essential?

dyongxu:
All checks passed. Merged this PR.

PR #504, opened 2023-01-11
feat: 🎸 [JIRA: HCPSDKFIORIUIKIT-1938] kpi header

xiaoyu0722:


Sources/FioriSwiftUICore/Experimental/KPIHeader.swift
```diff
@@ -1,217 +1,72 @@
...
+        }
+        
+        if data.count == 1 {
+            self.items = createOneItems().typeErased
+        } else if data.count == 2 {
+            self.items = createTwoItems().typeErased
+        } else if data.count == 3 {
+            self.items = createThreeItems().typeErased
+        } else if data.count >= 4 {
+            self.items = createFourItems().typeErased
```
billzhou0223, line 67: Can we simplify this code? 

```
KPIHeader {
ForEach(0..<min(data.count, 4)) {
    makeItem(data[$0])
}
}
```

I am not sure if limiting the number of items is necessary here. But if you do this way, you should document it properly.

billzhou0223 (changes requested):

billzhou0223 (approved):

PR #536, opened 2023-03-22
fix: 🐛 map fonts with unsupported weight and trait combo to existing 72 fonts.

billzhou0223:


MarcoEidinger:
@billzhou0223 the answer to your question" why we have to CI/build jobs" is that in `Package.swift` we specify `// swift-tools-version:5.5` and that implies that our Swift package shall be compilable with Xcode 13. An SDK developer may use accidentally language features from a different Swift version (e.g. 5.7 because developer is using Xcode 14). To detect and easily compare I introduced two CI/build jobs. Or are you asking we do have CI/buidls for `pull_request` and for `pull_request_target`?

I see the following build errors

```
Testing failed:
	Command CompileSwift failed with a nonzero exit code
	Value of type 'Font' has no member 'width'
	Cannot infer contextual base in reference to member 'condensed'
	Value of type 'Font' has no member 'width'
	Cannot infer contextual base in reference to member 'condensed'
	Value of type 'Font' has no member 'width'
	Cannot infer contextual base in reference to member 'condensed'
	Value of type 'Font' has no member 'width'
	Cannot infer contextual base in reference to member 'condensed'
```

MarcoEidinger:
Either we change swift-tools-version to 5.7 or you have to fix those erros

```swift
//font = font.width(.condensed)
font = font.width(Font.Width.condensed)
```

MarcoEidinger:
> I am talking about these two jobs. Looks like the error still persists.

Ignore the "pull_request_target" as this may take a different base commit and I wanted to remove that job anyway (originally I introduced it to allow code uploads to Sonar for pull requests coming from forks)

It is strange that the issue still persists for CI / build "pull_request". I see it uses your latest changes

Screenshot 2023-03-22 at 2 52 41 PM

But it complains about ` type 'Font' has no member 'Width'` which would be true < iOS 16 but you guard with a conditional statement 🤔

MarcoEidinger:
Issue reproducible when using Xcode 14.0 locally on my machine.

Build works fine in CI when using Xcode 14.2 (https://github.com/TeamEidinger/cloud-sdk-ios-fiori-1/actions/runs/4494999119/jobs/7908154011)

I assume the function was introduced later (16.1?)

billzhou0223:
> Issue reproducible when using Xcode 14.0 locally on my machine.
> 
> Build works fine in CI when using Xcode 14.2 (https://github.com/TeamEidinger/cloud-sdk-ios-fiori-1/actions/runs/4494999119/jobs/7908154011)
> 
> I assume the function was introduced later (16.1?)

Yeah it works for xcode 14.2. I tried to guard the code with both iOS 16.1 and 16.2 but neither worked.

MarcoEidinger:
Probably the best we can do is

```swift
#if swift(>=5.7.1)
  if #available(iOS 16.0, *) {
      font = font.width(Font.Width.condensed)
  }
#endif
```

https://forums.swift.org/t/using-apis-introduced-in-ios-minor-releases/63897/1

MarcoEidinger:
Alternatively, we can define `// swift-tools-version: 5.7.1` as the minimum which requires every consumer to use Xcode 14.1 while the previous solution allows lower Xcode versions but they won't benefit from your change.

billzhou0223:
> Probably the best we can do is
> 
> ```swift
> #if swift(>=5.7.1)
>   if #available(iOS 16.0, *) {
>       font = font.width(Font.Width.condensed)
>   }
> #endif
> ```
> 
> https://forums.swift.org/t/using-apis-introduced-in-ios-minor-releases/63897/1

I will take this approach. The impact should be minimal.

MarcoEidinger (approved):

PR #537, opened 2023-03-23
feat: 🎸 step progress indicator

xiaoyu0722:
|horizontal|vertical|
|-|-|
|Simulator Screen Shot - iPhone 14 Pro - 2023-03-23 at 17 31 04|Simulator Screen Shot - iPhone 14 Pro - 2023-03-23 at 17 31 07|


Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepProgressIndicator+View.swift
```diff
@@ -0,0 +1,277 @@
...
+        self.selection = selection
+        self.stepsData = stepsData
+        
+        self._currentStepName = EmptyView()
+        self._allStepsAction = EmptyView()
+        self._cancelAction = Action(model: _CancelActionDefault())
+        
+        self._steps = DefaultSteps(stepsData: stepsData,
+                                   selection: selection)
+    }
```
billzhou0223, line 115: In the default case, we only care about step title and the state, so a simple struct with two properties should be enough. 

```
public struct SingleStepItem {
    let title: String
    let state: State
}

init(selection: Binding<Int>, stepItems: [SingleStepItem]) {}
```

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepProgressIndicator+View.swift
```diff
@@ -0,0 +1,277 @@
...
+    {
+        self.selection = selection
+        self.stepsData = stepsData
+        self._currentStepName = currentStepName()
+        self._allStepsAction = EmptyView()
+        self._cancelAction = Action(model: _CancelActionDefault())
+        
+        self._steps = DefaultSteps(stepsData: stepsData,
+                                   selection: selection)
+    }
```
billzhou0223, line 161: We might not need these initializers. If developers want to provide their own views, they can just use the view builder based init generated by Sourcery.

billzhou0223 (changes requested):

billzhou0223:
> @billzhou0223
> 
> 1. Steps want individual state(normal, completed, disabled and error). These can combined with `selected` in a set.
>    But others are totally meaningless, like([normal, disabled, error]).
>    What steps really wanted may be:
> 
> ```swift
> State {
>  default(isSelected: Bool)
>  error(isSelected: Bool)
>  completed
>  disabled
> }
> ```
> 
> 2. For `SingleStepItem` just use title and state, I can't check if the it's a sub-step or just a normal one. I need the node.
> 3. And I have updated the codes, and use node as TextOrIcon.
>    But default StepModel still use node as a string.
>    If public `TextOrIcon` for default style. Developer may need handler the icon themselves. And I'm not sure if the `StepButtonStyle` will work as I wanted.
> 4. The selection statement `// sourcery: virtualPropSelection = "var selection: Binding<Int> = .constant(0)"`
>    I can't use `// sourcery: bindingProperty` in a view's protocol, `sourcery` seems can't generate code properly.
>    And the selection only be handled in default style.
>    When developer use `IndexedViewBuilder` to create steps. They should handle the selection and steps themselves?

1. You can use an OptionSet just like an enum. Using OptionSet give us the flexibility to support compound state in the future if requirement changes.

```
public struct StepState: OptionSet {
  public let rawValue: Int

  public init(rawValue: Int) {
    self.rawValue = rawValue
  }
}

extension StepState {
  public static let error = StepState(rawValue: 1 << 0)
  public static let disabled = StepState(rawValue: 1 << 1)
  public static let complete = StepState(rawValue: 1 << 2)
}

public struct StepItem {
  public let id: String
  public let title: String?
  public let state: StepState

  public init(id: String, title: String? = nil, state: StepState = []) {
    self.id = id
    self.title = title
    self.state = state
  }
}

let item = StepItem(id: UUID().uuidString)
var selection = item.id

switch item.state {
case .error:
  if item.id == selection {
    print("Give me a node for selected-error state")
  } else {
    print("Give me a node for error state")
  }
case .complete:
  print("Give me a node for complete state")
case .disabled:
  print("Give me a node for disabled state")
case []:
  if item.id == selection {
    print("Give me a node for selected state")
  } else {
    print("Give me a node for normal state")
  }
default:
  print("State not supported")
}
```

I think we need to change the type of selection from int to string (id) if we need to support sub-step use case. 

2. In order to support sub-step use cases, you can add a children property.

```
public struct StepItem {
 let id: String
 let title: String?
 let state: StepState
 let children: [StepItem]
}
```

You don't have to reply on `node` to tell if it's a sub step or not. Think the structure of step items as a tree. You can get the level of step items when you traverse the tree using DFS. For all the items at level 0, they are regular steps. Other items are sub steps.

3. `StepItem` will not need `node` property since SDK will generate the node for users based on the state.
4. Please checkout this doc first. https://github.com/SAP/cloud-sdk-ios-fiori/blob/main/CoreComponentGeneration.md#advanced-component-property-which-is-editable

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepProgressIndicator+View.swift
```diff
@@ -98,11 +98,11 @@ extension StepProgressIndicator: View {
 }
 
 public extension StepProgressIndicator {
-    init(selection: Binding<Int>,
-         stepsData: [StepModel])
+    init(selection: Binding<UUID>,
+         stepsData: [StepItem])
```
billzhou0223, line 102: Change param name to `stepItems`

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepsStyles.swift
```diff
@@ -26,10 +35,22 @@ enum InternalStepState {
...
 
-public protocol StepModel {
-    var title: String? { get }
-    var node: String? { get }
-    var state: StepIndicatorState { get }
+public struct StepItem: Identifiable {
+    public var id = UUID()
+    public var title: String?
+    public var state: StepIndicatorState
+    public var children: [StepItem]?
```
billzhou0223, line 42: Array doesn't need to be optional. You can assign an empty array as a default value to it.

billzhou0223 (changes requested):

billzhou0223 (changes requested):
Overall LGTM.

Sources/FioriSwiftUICore/Views/StepProgressIndicator/SingleStep.swift
```diff
@@ -12,12 +12,81 @@ extension Fiori {
...
+    /// Convenience initialization for empty title and sub-steps.
+    /// - Parameters:
+    ///   - stepId: String value for step id.
+    ///   - node: Node for single step.
+    init(stepId: String = UUID().uuidString,
+         @ViewBuilder node: () -> Node)
+    {
+        self.init(stepId: stepId, tappable: true, title: { EmptyView() }, node: node)
+    }
+}
```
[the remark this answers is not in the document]
xiaoyu0722: `Substeps` is not optional, `sourcery` did not handle this scenario.

billzhou0223 (approved):
Please add documentation later.

xiaoyu0722:
> Please add documentation later.

Doc is added.

PR #548, opened 2023-04-07
fix: incorrect color value for the elevatedContrastLight variant

CharlesXu0488:
Problem was found in processing input parameters mapping to variant.
Fix is to add cases of switch statement for the missing cases.


billzhou0223 (commented):
Please fix CI failures

billzhou0223 (approved):

PR #563, opened 2023-05-10
fix: bcp 2370057173 chart8 and chart9 color mismatch

CharlesXu0488:
update chart8 and chart9 colors


billzhou0223 (approved):
Please make sure your commit message follows conventional commit standard to prevent CI check failure.

PR #577, opened 2023-06-05
fix: 🐛 searchable list voiceover actions failed

xiaoyu0722:


billzhou0223 (changes requested):
What do you mean by sharedAction? In which use case this new api is needed?

billzhou0223 (changes requested):
What do you mean by sharedAction? In which use case this new api is needed?

xiaoyu0722:
> 

`sharedAction` is a local action for `Action` that we can handle the action without tap gesture.
It can be triggered same as `_didSelectAction`.
Then voiceover will trigger the action.

billzhou0223:
> > 
> 
> `sharedAction` is a local action for `Action` that we can handle the action without tap gesture. It can be triggered same as `_didSelectAction`. Then voiceover will trigger the action.

I wouldn't use an environment variable for this purpose. You can add a property in `Action` so that is is only visible in its own scope instead of exposing it in the environment.

xiaoyu0722:
> > > 
> > 
> > 
> > `sharedAction` is a local action for `Action` that we can handle the action without tap gesture. It can be triggered same as `_didSelectAction`. Then voiceover will trigger the action.
> 
> I wouldn't use an environment variable for this purpose. You can add a property in `Action` so that is is only visible in its own scope instead of exposing it in the environment.

The setting is missed, the action is actually a view wrapped into `some view`.

PR #578, opened 2023-06-05
fix: 🐛 duration picker voiceover issue

xiaoyu0722:


billzhou0223 (changes requested):
The reason your previous implementation has issue is because you created the picker view as a global variable. Instead you should create it as a local variable in `makeUIView` and return it.

xiaoyu0722:
> The reason your previous implementation has issue is because you created the picker view as a global variable. Instead you should create it as a local variable in `makeUIView` and return it.

For duration picker, yes, I use the local picker view in makeUIView. No logic changes, just move them to 
DurationPickerContainer. This will make accessibility works fine.
```swift
        let pickerView = DurationPickerContainer(selection: $selection,
                                                 maximumMinutes: maximumMinutes,
                                                 minimumMinutes: minimumMinutes,
                                                 minuteInterval: minuteInterval,
                                                 measurementFormatter: measurementFormatter)
        pickerView.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
        pickerView.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
        return pickerView
```

billzhou0223:
> > The reason your previous implementation has issue is because you created the picker view as a global variable. Instead you should create it as a local variable in `makeUIView` and return it.
> 
> For duration picker, yes, I use the local picker view in makeUIView. No logic changes, just move them to DurationPickerContainer. This will make accessibility works fine.
> 
> ```swift
>         let pickerView = DurationPickerContainer(selection: $selection,
>                                                  maximumMinutes: maximumMinutes,
>                                                  minimumMinutes: minimumMinutes,
>                                                  minuteInterval: minuteInterval,
>                                                  measurementFormatter: measurementFormatter)
>         pickerView.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
>         pickerView.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
>         return pickerView
> ```

`DurationPickerViewWrapper` and `DurationPickerContainer` really confuse me if I just look at their names without looking into the implementation. At least you should rename them to make them self-explanatory. TBH I don't think adding another container class is necessary which complicates the overall hierarchy.

xiaoyu0722:
I have revert it, and use local picker view.
But still need a container for picker view, or accessibility still wrong. I don't know why. 🤣

billzhou0223 (approved):

PR #591, opened 2023-06-21
chore: [HCPSDKFIORIUIKIT-2283] Upgrade swift tools version to 5.7

angiexyang:


billzhou0223:
@MarcoEidinger Please advice if we need to update the xcode and tooling version.

MarcoEidinger:
@billzhou0223 it looks to me that the respective files

- `Package.swift` for tools version (which has an impact on the Xcode version to be used)
- `ci.yml` to deprecate build (macos-11, Xcode_13.0)) 

and documentation (`README.md`) was updated.

Or do you have something else in mind when you say "to update the xcode and tooling version." ?

There are probably more conditional compilation statements (e.g. `#if swift(>=5.5)`) which can be removed.

Later we need to change protection branch rule to make action to `build (macos-12, Xcode_14.0)` required but that is repository configuration and not part of PR.

billzhou0223:
> @billzhou0223 it looks to me that the respective files
> 
> * `Package.swift` for tools version (which has an impact on the Xcode version to be used)
> * `ci.yml` to deprecate build (macos-11, Xcode_13.0))
> 
> and documentation (`README.md`) was updated.
> 
> Or do you have something else in mind when you say "to update the xcode and tooling version." ?
> 
> There are probably more conditional compilation statements (e.g. `#if swift(>=5.5)`) which can be removed.
> 
> Later we need to change protection branch rule to make action to `build (macos-12, Xcode_14.0)` required but that is repository configuration and not part of PR.

I didn't realize this update is coming from a JIRA item proposed by you.. so I am a little confused in the first place without knowing the purpose of the change but now I think we are good.

billzhou0223 (approved):

PR #602, opened 2023-10-16
fix: 🐛 [BCP:2370082757] Fix SignatureCaptureView VO issues

janhuachu:
✅ Closes: 2370082757

### DIFF
--- a/Sources/FioriSwiftUICore/Views/Action+View.swift
+++ b/Sources/FioriSwiftUICore/Views/Action+View.swift
@@ -21,6 +21,9 @@ extension Action: View {
             } label: {
                 Text(self._actionText ?? "")
             }
+            .accessibilityElement()
+            .accessibilityLabel(_actionText ?? "")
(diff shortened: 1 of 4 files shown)


Sources/FioriSwiftUICore/Views/Action+View.swift
```diff
@@ -21,6 +21,9 @@ extension Action: View {
             } label: {
                 Text(self._actionText ?? "")
             }
+            .accessibilityElement()
```
billzhou0223, line 24: Is this modifier needed?

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/Action+View.swift
janhuachu: Yes, it is needed. Without this line, the VO is fine but double tap will not trigger action. Strange.

billzhou0223 (approved):

PR #605, opened 2023-10-18
feat: 🎸 [JIRA: HCPSDKFIORIUIKIT-2444] Update to include more icons

JuliaSchnSAP:


billzhou0223 (commented):
We used to put icons in the same category in a dedicated folder to form this folder hierarchy. 

image

Just curious what the reason is to abort this pattern and use a flat list instead?

JuliaSchnSAP:
> We used to put icons in the same category in a dedicated folder to form this folder hierarchy.
> 
> image
> 
> Just curious what the reason is to abort this pattern and use a flat list instead?

I can change it back to a folder structure. There are some icons that don't exist in any category, so how do you suggest that I handle them? I was thinking that I could put all the icons under FioriSFSymbols > All or use the previous folder structure with an Uncategorized folder

billzhou0223 (approved):

billzhou0223:
Please fix CI failures (`PR / CodeFormattingCheck` can be ignored for now.)

PR #621, opened 2023-12-01
feat: 🎸 [JIRA: HCPSDKFIORIUIKIT-2459] Tab Bar

xiaoyu0722:


billzhou0223 (commented):
Do we need to provide default styles for tab bar items similar to what we did for nav bar?

Apps/Examples/Examples/FioriSwiftUICore/TabViewExample/TabViewDetailView.swift
```diff
@@ -0,0 +1,106 @@
...
+    var imageName: String
+}
+
+struct TabViewDetailView: View {
+    @Binding var numberOfTabs: Int
+    @Binding var isCustomColor: Bool
+    @Binding var isCustomSelectionIndicatorImage: Bool
+    @Binding var isCustomBackgroundImage: Bool
+    @State private var selection = 0
+    @State private var accendColor = Color.orange
```
dyongxu, line 16: Change the color to use Fiori based color

Apps/Examples/Examples/FioriSwiftUICore/TabViewExample/TabViewDetailView.swift
```diff
@@ -0,0 +1,106 @@
...
+                        icon: { Image(systemName: t.imageName) }
+                    )
+                }
+            }
+        }
+        .navigationTitle("Customized TabView")
+        .accentColor(accendColor)
+        .onAppear {
+            if isCustomColor {
+                UITabBarItem.appearance().setTitleTextAttributes([.font: UIFont.preferredFioriFont(forTextStyle: .title3),
```
dyongxu, line 63: Change the color to use Fiori based color, for the codes in this example.

dyongxu (changes requested):
1. Try to use Fiori Color , wherever possible
2. Add the support for tab haptic sound. In SDK UIKit, when the user taps the tab, it will have vibration and sound.

Apps/Examples/Examples/FioriSwiftUICore/TabViewExample/TabViewDetailView.swift
dyongxu: There are lots of places using the native color from Apple in this PR. Let's try to switch to Fiori Color as much as possible.

xiaoyu0722:
> Do we need to provide default styles for tab bar items similar to what we did for nav bar?

All this setting are for test app, also I will restore to system default setting when view are disappeared.

dyongxu (approved):

dyongxu:
> > Do we need to provide default styles for tab bar items similar to what we did for nav bar?
> 
> All these settings are for test app, also I will restore to system default setting when view are disappeared.

Yes, it will be a good idea to use Fiori Color and Font, even if in the TestApp, just to show Fiori Styles can be actually applied.

PR #622, opened 2023-12-15
feat: merge visionOS branch into main

dyongxu:


billzhou0223 (approved):

Package@swift-5.9.swift
billzhou0223, line 1: What's the reason to add this new file rather than updating the existing `package.swift` file?

Usually we just need to update the version number in `package.swift`.

billzhou0223 (changes requested):

Package@swift-5.9.swift
dyongxu: @MarcoEidinger  Shall we merge the contents from Package@swift-5.9.swift to Package.swift?  Looks like , this "new" file is no longer needed in main branch ?

MarcoEidinger:
@billzhou0223  and @dyongxu : an important question: do you want to expose all modules with their APIs for visionOS even `SAPFioriSwiftUICore` and `SAPFioriCharts` were not optimized and tested? If not, then additional efforts are needed, either excluding modules in `Package.swift` or excluding individual APIs with conditional compilation statements.

dyongxu:
I am ok with exposing all APIs, with a "disclaimer" that not all UIs for visionOS are optimized for native experience. @MarcoEidinger @billzhou0223

billzhou0223:
> I am ok with exposing all APIs, with a "disclaimer" that not all UIs for visionOS are optimized for native experience. @MarcoEidinger @billzhou0223

LGTM.

MarcoEidinger:
**DO NOT DELETE** visionOS branch as it is still referenced by SAP teams

MarcoEidinger (approved):

PR #624, opened 2024-01-13
feat: [JIRA: HCPSDKFIORIUIKIT-2452] Empty State View: Illustration and Message Enhancements

JuliaSchnSAP:


Sources/FioriSwiftUICore/_generated/ViewModels/API/EmptyStateView+API.generated.swift
```diff
@@ -23,12 +24,14 @@ public struct EmptyStateView<Title: View, DescriptionText: View, DetailImage: Vi
         @ViewBuilder title: () -> Title,
 		@ViewBuilder descriptionText: () -> DescriptionText,
 		@ViewBuilder detailImage: () -> DetailImage,
-		@ViewBuilder action: () -> ActionView
+		@ViewBuilder action: () -> ActionView,
+        alignmentAxis: Axis = .vertical
```
billzhou0223, line 28: files ended with `.generated` cannot be edited manually otherwise the changes will get lost when sourcery runs next time. You should update the model definition then run sourcery program to generate the api for you.

Where does this `alignmentAxis` requirement come from? I am not seeing it on the spec.

billzhou0223 (changes requested):

PR #645, opened 2024-02-22
feat: 🎸 [JIRA:0] TextInputInfoView for TextInput

xiaoqinggrace:


Sources/FioriSwiftUICore/_FioriStyles/TextInputInfoViewStyle.fiori.swift
```diff
@@ -5,7 +5,7 @@ import SwiftUI
 // Base Layout style
 public struct TextInputInfoViewBaseStyle: TextInputInfoViewStyle {
     public func makeBody(_ configuration: TextInputInfoViewConfiguration) -> some View {
-        GeometryReader { geometry in
+        GeometryReader { _ in
             VStack {
```
janhuachu, line 9: GeometryReader is not used. Should be removed.

Sources/FioriSwiftUICore/_FioriStyles/TextInputInfoViewStyle.fiori.swift
```diff
@@ -14,9 +14,6 @@ public struct TextInputInfoViewBaseStyle: TextInputInfoViewStyle {
                     Spacer()
```
janhuachu, line 14: The Spacer and counter is still there that would put some space between the description and the end of the trailing edge. I think it should be like

`                if !configuration.counter.isEmpty {
                    Spacer()
                    configuration.counter
                }`

janhuachu (approved):

PR #655, opened 2024-03-08
feat: 🎸 [HCPSDKFIORIUIKIT-2452] Add vertical layout style modifier to Illustrated Message

JuliaSchnSAP:
This component has a vertical layout by default. Adding a style modifier for this layout orientation would allow users to explicitly select either layout style instead of just the horizontal style. This would make the API more consistent

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/IllustratedMessageStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/IllustratedMessageStyle.fiori.swift
@@ -1,0 +1,88 @@
+import FioriThemeManager
+import Foundation
+import SwiftUI
+
+/**
+ This file provides default fiori style for the component.
+ 
+ 1. Uncomment fhe following code.
+ 2. Implement layout and style in corresponding places.
+ 3. Delete `.generated` from file name.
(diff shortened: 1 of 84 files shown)


Sources/FioriSwiftUICore/_FioriStyles/IllustratedMessageStyle.fiori.swift
```diff
@@ -62,9 +62,11 @@ extension IllustratedMessageFioriStyle {
     }
 }
 
+/// Vertical layout style of the Illustrated Message. All content is displayed in one row. This is the default layout style
```
dyongxu, line 65: Do you actually mean one Column?

Sources/FioriSwiftUICore/_FioriStyles/IllustratedMessageStyle.fiori.swift
```diff
@@ -80,6 +82,13 @@ public struct IllustratedMessageHorizontalLayoutStyle: IllustratedMessageStyle {
...
```
dyongxu, line 85: is it actually one column?

dyongxu (changes requested):
may update API doc

billzhou0223 (approved):

PR #665, opened 2024-03-16
feat: 🎸 [HCPSDKFIORIUIKIT-2481] New Mobile Card

shengxu7:


Sources/FioriSwiftUICore/Views/TagStack/Tag.swift
```diff
@@ -59,6 +59,28 @@ public struct DarkTagStyle: TagStyle {
     }
 }
 
+public struct ColorTagStyle: TagStyle {
```
billzhou0223, line 62: It's recommended to add a static method in the tag style extension for convenient access to this type of style. I will also separate foreground color and fill color

```
public extension TagStyle where Self == ForegroundColorTagStyle {
    static func foregroundColor(_ color: Color) -> ForegroundColorTagStyle {
        ForegroundColorTagStyle(color)
    }
}

public extension TagStyle where Self == FillColorTagStyle {
    static func fillColor(_ color: Color) -> FillColorTagStyle {
        FillColorTagStyle(color)
    }
}
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/TagStack/Tag.swift
shengxu7: Removed this public style.

billzhou0223 (approved):

PR #669, opened 2024-03-21
feat: 🎸 [JRIA:HCPSDKFIORIUIKIT-2460] TextInput form cells

janhuachu:
TitleFormView and TextFieldFormView implementation

✅ Closes: HCPSDKFIORIUIKIT-2460

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
@@ -1,0 +1,154 @@
+import FioriThemeManager
+import Foundation
+import SwiftUI
+
+/// The base layout style for `TextFieldFormView`.
+public struct TextFieldFormViewBaseStyle: TextFieldFormViewStyle {
+    public func makeBody(_ configuration: TextFieldFormViewConfiguration) -> some View {
+        VStack(alignment: .leading) {
+            configuration.title
+            configuration._titleFormView
+        }
+    }
+}
+    
+// Default fiori styles
+extension TextFieldFormViewFioriStyle {
+    struct ContentFioriStyle: TextFieldFormViewStyle {
+        @FocusState var isFocused: Bool
+        
+        func makeBody(_ configuration: TextFieldFormViewConfiguration) -> some View {
+            TextFieldFormView(configuration)
+                .titleStyle { titleConf in
+                    Title(titleConf)
+                        .foregroundStyle(self.getTitleColor(configuration))
+                        .font(.fiori(forTextStyle: .subheadline, weight: .semibold))
+                }
+                .placeholderTextFieldStyle { config in
+                    HStack {
+                        PlaceholderTextField(config)
+                            .foregroundStyle(self.getTextColor(configuration))
+                            .font(.fiori(forTextStyle: .body))
+                            .accentColor(self.getCursorColor(configuration))
+                            .focused(self.$isFocused)
+                            .onChange(of: configuration.text) { s in
+                                self.checkCharCount(configuration, textString: s)
+                            }
+                            .frame(minHeight: 44)
+                            .padding(EdgeInsets(top: 0, leading: 8, bottom: 0, trailing: 8))
+                        if self.showsActionButton(configuration) {
+                            Button(action: self.getAction(configuration)!) {
+                                self.getActionIcon(configuration)!
+                            }
+                            .padding(.trailing, 8)
+                        }
+                    }
+                    .background(RoundedRectangle(cornerRadius: 8).stroke(self.getBorderColor(configuration), lineWidth: self.getBorderWidth(configuration)).background(self.getBackgroundColor(configuration)))
+                    .cornerRadius(8)
+                }
+                .textInputInfoViewStyle { config in
+                    let isError = self.isErrorStyle(configuration)
+                    let style: any TextInputInfoViewStyle = isError ? .error : .fiori
+                    TextInputInfoView(config)
+                        .textInputInfoViewStyle(style)
+                        .typeErased
+                }
+        }
+
+        func isDisabled(_ configuration: TextFieldFormViewConfiguration) -> Bool {
+            configuration.controlState == .disabled
+        }
+
+        func getTitleColor(_ configuration: TextFieldFormViewConfiguration) -> Color {
+            guard !self.isDisabled(configuration) else {
+                return .preferredColor(.separator)
+            }
+            return self.isErrorStyle(configuration) ? .preferredColor(.negativeLabel) : .preferredColor(.primaryLabel)
+        }
+
+        func getTextColor(_ configuration: TextFieldFormViewConfiguration) -> Color {
+            TextInputFormViewConfiguration(configuration, isFocused: self.isFocused).getTextColor()
+        }
+
+        func getCursorColor(_ configuration: TextFieldFormViewConfiguration) -> Color {
+            TextInputFormViewConfiguration(configuration, isFocused: self.isFocused).getCursorColor()
+        }
+
+        func getBackgroundColor(_ configuration: TextFieldFormViewConfiguration) -> Color {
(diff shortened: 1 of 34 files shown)


janhuachu:
Also fixed NoteFormView and KeyValueFormView issues.

Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
```diff
@@ -0,0 +1,154 @@
...
+}
+    
+// Default fiori styles
+extension TextFieldFormViewFioriStyle {
+    struct ContentFioriStyle: TextFieldFormViewStyle {
+        @FocusState var isFocused: Bool
+        
+        func makeBody(_ configuration: TextFieldFormViewConfiguration) -> some View {
+            TextFieldFormView(configuration)
+                .titleStyle { titleConf in
```
billzhou0223, line 22: This style can be set in TextFieldFormViewFioriStyle.TitleFioriStyle.

Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
```diff
@@ -0,0 +1,154 @@
...
+                    Title(titleConf)
+                        .foregroundStyle(self.getTitleColor(configuration))
+                        .font(.fiori(forTextStyle: .subheadline, weight: .semibold))
+                }
+                .placeholderTextFieldStyle { config in
```
billzhou0223, line 27: Can we move it into the PlaceholderStyle section?

Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
```diff
@@ -0,0 +1,154 @@
...
+                            Button(action: action) {
+                                actionIcon
+                            }
+                            .padding(.trailing, 8)
+                        }
+                    }
+                    .background(RoundedRectangle(cornerRadius: 8).stroke(self.getBorderColor(configuration), lineWidth: self.getBorderWidth(configuration)).background(self.getBackgroundColor(configuration)))
+                    .cornerRadius(8)
+                }
+                .textInputInfoViewStyle { config in
```
billzhou0223, line 49: Can we move it into the TitleInputInfoViewStyle section?

Sources/FioriSwiftUICore/_FioriStyles/TextViewStyle.fiori.swift
```diff
@@ -24,7 +15,5 @@ public struct TextViewFioriStyle: TextViewStyle {
     @ViewBuilder
     public func makeBody(_ configuration: TextViewConfiguration) -> some View {
         TextView(configuration)
-            .foregroundStyle(Color.preferredColor(.primaryLabel))
-            .font(.fiori(forTextStyle: .body))
```
billzhou0223, line 28: Why do you remove them? I think text view needs to have some default styles.

Sources/FioriSwiftUICore/_FioriStyles/TitleFormViewStyle.fiori.swift
```diff
@@ -0,0 +1,120 @@
...
+        @FocusState var isFocused: Bool
+
+        func makeBody(_ configuration: TitleFormViewConfiguration) -> some View {
+            VStack {
+                TitleFormView(configuration)
+                    .foregroundStyle(self.getTextColor(configuration))
+                    .font(.fiori(forTextStyle: .body))
+                    .accentColor(self.getCursorColor(configuration))
+                    .focused(self.$isFocused)
+                    .textInputInfoViewStyle { config in
```
billzhou0223, line 48: Can we move it into the corresponding style definition instead of in the content style?

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
janhuachu: The color of the title text depends on the controlState of the TextFieldFormViewConfiguration.

In makeBody of the TextFieldFormViewFioriStyle.TitleFioriStyle it has no reference to that controlState

Sources/FioriSwiftUICore/_FioriStyles/TextViewStyle.fiori.swift
janhuachu: If I don't remove them, the text is always has .promaryLable. I have not found a way to set it to the color when the controlState is disabled.

Any idea how I can override this color in TextFielFormViewStyle?

Sources/FioriSwiftUICore/_FioriStyles/TitleFormViewStyle.fiori.swift
janhuachu: Same that we need the controlState to determine the colors while controlState is only in the TitleFormViewConfiguration.

janhuachu:
@billzhou0223 Is it possible to have the subcomponent has the same property in its super-component?
In this case, the controlState is needed in TextInputFieldComponent, PlaceholderTextFieldComponent, and TitleFormViewComponent?

Sources/FioriSwiftUICore/_FioriStyles/TextViewStyle.fiori.swift
janhuachu: This is the same problem that the controlState of the NoteFormViewConfiguration is not available in NoteFormViewFioriStyle.TextViewFioriStyle

billzhou0223: My understanding is that `NoteFormViewFioriStyle.TextViewFioriStyle` should overwrite `TextViewFioriStyle` so that you don't need to remove these default styles here.

janhuachu: Yes. However, NoteFormViewFioriStyle.TextViewFioriStyle does not have access to NoteFormViewConfiguration which has the controlState property

billzhou0223: OK. You can leave the text view styles in `NoteFormViewFioriStyle.ContentStyle`. But you don't need to remove these default style here right because they will be overwritten by `NoteFormViewFioriStyle.ContentStyle`

billzhou0223:
> @billzhou0223 Is it possible to have the subcomponent has the same property in its super-component? In this case, the controlState is needed in TextInputFieldComponent, PlaceholderTextFieldComponent, and TitleFormViewComponent?

I see what you mean but it's not possible based on how the style protocol is defined. If the style of sub-components needs to rely on super-components, you current implementation should be the correct way to go.

Sources/FioriSwiftUICore/_FioriStyles/TextViewStyle.fiori.swift
janhuachu: I have to remove it. Otherwise  Below is the sequence. As observed that somehow step 7 is too late to be effective.

1. PlaceholderTextEditorFioriStyle.ContentFioriStyle (no controlState available)
2. NoteFormViewFioriStyle.PlaceholderTextEditorFioriStyle (no controlState available)
3. NoteFormViewFioriStyle.ContentFioriStyle (SET TO DISABLED COLOR)
4. TextViewFioriStyle (SET TO DEFAULT COLOR)
5. PlaceholderTextEditorFioriStyle.TextViewFioriStyle (no controlState available)
6. NoteFormViewFioriStyle.TextViewFioriStyle (no controlState available)

7. NoteFormViewFioriStyle.ContentFioriStyle (Try to change to disabled color but the view is already displayed. Has no effect.)

billzhou0223 (approved):

PR #674, opened 2024-03-28
feat: 🎸 [JIRA: HCPSDKFIORIUIKIT-2559] list style configurable

xiaoyu0722:
**Requirement:**
I would like to be able to configure the ListPicker's ListpickerItemConfiguration's destinationView's List liststlyle. The current implementation of the ListPicker uses the default liststyle .insetGrouped. Our AC to have the destinationView's with a grouped liststyle, but it would be nice to have a way to configure the style for ourselves.


Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
```diff
@@ -234,3 +233,24 @@ extension ListPickerItem {
         }
     }
 }
+
+struct DestinationListStyleKey: EnvironmentKey {
```
billzhou0223, line 237: Rename to `PickerListStyle`

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
```diff
@@ -234,3 +233,24 @@ extension ListPickerItem {
...
+        get { self[DestinationListStyleKey.self] }
+        set { self[DestinationListStyleKey.self] = newValue }
+    }
+}
+
+public extension View {
+    /// List style for destination list in list picker.
+    /// - Parameter style: some `ListStyle`.
+    /// - Returns: New destination list style for list picker.
+    func destinationStyle(_ style: any ListStyle) -> some View {
```
billzhou0223, line 253: listPickerListStyle

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
xiaoyu0722: I tried to use the same name as its environment value, but can be compiled. I don't know why, just change a name, everything be good.
Screenshot 2024-03-29 at 13 42 52

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
xiaoyu0722: Developer can set style for list picker directly, `ListPicker {}.listStyle(.grouped)`
Currently only destination view need to be styleable.

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
billzhou0223: Try to change the parameter type from `any ListStyle` to `some ListStyle`

Sources/FioriSwiftUICore/Views/ListPickerItem+View.swift
billzhou0223: I want to avoid using destination as a name of this API because it's a parameter name of navigation link and sdk users may not understand what it is used for if they don't know we are using navigation link internally. 

`ListPicker {}.listStyle(.grouped)` is not a valid use case since ListPicker itself is a just a view not a list. Setting listStyle won't have any affect on the destination list view, that 's why we need an additional API for it.

xiaoyu0722: @billzhou0223 Or can we use `sublistStyle`?

xiaoyu0722:
Still can not use same name, I use `listPickerListStyle` for public modifier and `listpickerListStyle` for internal environment value.

billzhou0223 (approved):

PR #681, opened 2024-04-19
feat: 🎸 [HCPSDKFIORIUIKIT-2601]Add additional color style to V8

dyongxu:
1. As per Design, added 6 new color styles for Joule Mobile Client
2. Updated unit test for V8
3. Added UI tests for Gradient Color effect ( borderline, background )

### DIFF
--- a/Apps/Examples/Examples/FioriThemeManager/Colors.swift
+++ b/Apps/Examples/Examples/FioriThemeManager/Colors.swift
@@ -1,3 +1,8 @@
+🌱 Cloning SwiftFormat 0.53.4
+🌱 Resolving package
+🌱 Building product swiftformat
+🌱 Installed SwiftFormat 0.53.4
+🌱 Running swiftformat 0.53.4...
 import FioriThemeManager
 import SwiftUI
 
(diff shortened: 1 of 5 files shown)


Apps/Examples/Examples/FioriThemeManager/Colors.swift
```diff
@@ -1,3 +1,8 @@
+🌱 Cloning SwiftFormat 0.53.4
+🌱 Resolving package
+🌱 Building product swiftformat
+🌱 Installed SwiftFormat 0.53.4
+🌱 Running swiftformat 0.53.4...
```
billzhou0223, line 5: There lines should be removed.

billzhou0223 (changes requested):

Apps/Examples/Examples/FioriThemeManager/Colors.swift
dyongxu: Removed. But did not find out where they came from. To be verified again after CI build is complete.

dyongxu: It is good now.

billzhou0223 (approved):

PR #683, opened 2024-04-25
feat: 🎸 list picker enhancement

xiaoyu0722:
Jira: IOSSDKBUG-156 SwiftUI ListPickerItem can't add toolbar
## Caution: Public API changed
`.listPickerListStyle` is changed to a new one to support customization for destination list in list picker..
|before|after|
|-|-|
|Simulator Screenshot - iPhone 15 Pro Max - 2024-04-25 at 15 58 44|Simulator Screenshot - iPhone 15 Pro Max - 2024-04-25 at 15 57 51|

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPicker.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPicker.swift
@@ -127,7 +127,17 @@ struct ListPickerItemDataNonIdentifiableExample: View {
                 Text(framework.name)
             }))
         }
-        .listPickerListStyle(.inset)
+        .listPickerListView { c in
+            c.toolbar {
+                ToolbarItem(placement: .topBarTrailing) {
+                    Button {
+                        print("Tool bar item tapped.")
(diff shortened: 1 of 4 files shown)


Sources/FioriSwiftUICore/Views/ListPicker/ListPicker+Styleable.swift
billzhou0223, line 1: No need to use this design from styleable component. In the future, you can recreate ListPicker using the new pattern and let Sourcery generate the code for you.

If your goal is to let developers able to customize the list view, simply exposing the destination view through an view modifier is good enough (take `.titleModifier() as an example`).

```
public extension View {

    @ViewBuilder
    func listPickerListViewModifier<V: View>(_ transform: @escaping (AnyViewModifier.Content) -> V) -> some View {
        self.environment(\.listPickerListViewModifier, AnyViewModifier(transform))
    }
```

billzhou0223 (changes requested):

billzhou0223 (approved):

PR #692, opened 2024-05-20
feat: 🎸 [JIRA:IOSSDKBUG-5] add variables to ChartLabelAttribute

hengyi-zhang:
add two variables(font and backgroundColor) to ChartLabelAttributes, for user to custom chart label attributes conveniently


Sources/FioriCharts/Charts/Common/AxisTitle.swift
```diff
@@ -1,3 +1,8 @@
+🌱 Cloning SwiftFormat 0.53.4
+🌱 Resolving package
+🌱 Building product swiftformat
+🌱 Installed SwiftFormat 0.53.4
+🌱 Running swiftformat 0.53.4...
```
billzhou0223, line 5: Remove these redundant texts.

Sources/FioriCharts/Charts/Common/XAxisView.swift
```diff
@@ -51,8 +51,8 @@ struct XAxisView: View {
                 ForEach(0 ..< labels.count, id: \.self) { index in
                     Group {
                         if isShowLabels[index] {
-                            Text(labels[index].title)
-                                .font(.fiori(fixedSize: axis.labels.fontSize))
+                            Text(labels[index].attributedTitle(axis.labels.mergedAttributes))
+                                .font(axis.labels.font ?? .fiori(fixedSize: axis.labels.fontSize))
```
billzhou0223, line 55: Do you need to set font here if the font attribute is already included in attributed string?

Sources/FioriCharts/Charts/Common/AxisTitle.swift
```diff
@@ -36,3 +41,10 @@ struct AxisTitle: Identifiable, Hashable {
         hasher.combine(self.size.height)
     }
 }
+
+extension AxisTitle {
+    func attributedTitle(_ styleSheetAttributes: [NSAttributedString.Key: Any]) -> AttributedString {
```
billzhou0223, line 46: What is `styleSheetAttributes`? We haven't added component-level stylesheet support in SwiftUI yet.

Sources/FioriCharts/Model/ChartLabelAttributes.swift
```diff
@@ -17,24 +18,58 @@ public class ChartLabelAttributes: ObservableObject, Identifiable, NSCopying {
...
+    /// Font for the label, has higher priority than fontSize.
+    @Published public var font: Font?
+    /// background color for the label.
+    @Published public var backgroundColor: Color?
+    
+    /// Style attributes for the label.
+    @Published public var styleAttributes: [NSAttributedString.Key: Any] = [:]
+    
+    /// merge current style(backgroundColor and foregroundColor) and  styleAttributes
+    var mergedAttributes: [NSAttributedString.Key: Any] {
```
billzhou0223, line 30: 1. Why do we need to have both public style properties like `font`, `backgroundColor` and `styleAttributes`? These two types of API seem to have the same purpose so consider only keeping one of them. 
2. Do we want to reuse the same design in SAPFiori where we separate attributes from different sources (e.g. default, developer-set, stylesheet) in its own dictionary so that we can merge them in a certain order at rendering time?

@shengxu7 Your thoughts?

billzhou0223 (changes requested):

Sources/FioriCharts/Model/ChartLabelAttributes.swift
shengxu7: Don't add styleAttributes at this moment.

Sources/FioriCharts/Charts/Common/AxisTitle.swift
hengyi-zhang: fixed!

Sources/FioriCharts/Charts/Common/XAxisView.swift
hengyi-zhang: attributed string has been removed

Sources/FioriCharts/Model/ChartLabelAttributes.swift
hengyi-zhang: fixed!

Sources/FioriCharts/Charts/Common/AxisTitle.swift
hengyi-zhang: fixed

shengxu7 (approved):

shengxu7 (approved):

billzhou0223 (approved):

PR #701, opened 2024-06-07
feat: 🎸 [JIRA-HCPSDKFIORIUIKIT-2606] Form Cells: Mandator Field

xiaoqinggrace:
feat: [JIRA-HCPSDKFIORIUIKIT-2606] Form Cells: Mandatory Field Indicator


Apps/Examples/Examples/FioriSwiftUICore/SignatureView/SignatureCaptureViewExample.swift
```diff
@@ -43,6 +43,7 @@ struct SignatureCaptureViewExample2: View {
                              .hidesXmark(false)
                              .hidesSignatureLine(true)
                              .addsTimestampInImage(true)
+                             .isRequired(true)
```
dyongxu, line 46: Add a switch control , similar to the other two test cases updated,  to turn on/off the "required indicator".

dyongxu (changes requested):
Add a switch control

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
```diff
@@ -508,6 +538,18 @@ public extension SignatureCaptureView {
...
+     A view modifier to indicate if the component is a mandatory field.
+
+     - parameter `isRequired`: A boolean variable to indicate if the cell is a mandatory field.
+     The default value is `false`.
+     */
+    func isRequired(_ isRequired: Bool) -> Self {
+        var newSelf = self
+        newSelf.isRequired = isRequired
+        return newSelf
+    }
```
billzhou0223, line 552: We should make `isRequired` a environment value because it is a common property across all form views. So that you don't need to add it into each component individually.

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
```diff
@@ -248,7 +249,36 @@ extension SignatureCaptureView: View {
...
+        }
+        if isRequired {
+            if titleString?.suffix(1) != "*" {
+                return (titleString ?? "") + "*"
+            }
+        } else {
+            if titleString?.suffix(1) == "*" {
+                return String(titleString?.dropLast(1) ?? "")
+            }
+        }
```
billzhou0223, line 266: This logic is very simple. If isRequired, append *. We don't need to handle else case.

Sources/FioriSwiftUICore/_ComponentProtocols/BaseComponentProtocols.swift
```diff
@@ -17,6 +17,13 @@ protocol _SubtitleComponent {
     var subtitle: AttributedString? { get }
 }
 
+// sourcery: BaseComponent
+protocol _MandatoryIndicatorComponent {
```
billzhou0223, line 21: I remember you said we don't want developers to customize the indicator so only exposing `isRequired` should be enough right?

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -84,7 +84,11 @@ protocol _NoteFormViewComponent: _PlaceholderTextEditorComponent, _FormViewCompo
 }
 
 // sourcery: CompositeComponent
-protocol _KeyValueFormViewComponent: _TitleComponent, _NoteFormViewComponent {}
+protocol _KeyValueFormViewComponent: _TitleComponent, _NoteFormViewComponent, _MandatoryIndicatorComponent {
+    // sourcery: defaultValue = false
+    ///  Indicates whether the cell is a mandatory field. The default value is `false`.
+    var isRequired: Bool { get }
```
billzhou0223, line 90: Make `isRequired` a environment value instead of adding it in every component.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
xiaoqinggrace: Based on UX spec, the mandatory field is the feature for several form cells, not for all components.

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
xiaoqinggrace: There might be a use case in which the field is mandatory based on certain conditions such as the value of another field. I think it will be. more flexible by providing both if and else handling.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
xiaoqinggrace: Based on UX spec, the mandatory field is the feature for several form cells, not for all components.

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
xiaoqinggrace: Updated

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
xiaoqinggrace: The environment value defined in the base component model is not found in the generated API.

billzhou0223 (changes requested):

billzhou0223 (approved):

PR #705, opened 2024-06-12
feat: MenuSelection implementation

billzhou0223:


Sources/FioriSwiftUICore/Utils/AttributedString+Extensions.swift
shengxu7, line 1: formatKey is not used.

shengxu7 (approved):

PR #711, opened 2024-06-13
chore: 🤖[HCPSDKFIORIUIKIT-2608] Clean up warnings : implicit conformance 

KevinZK:
# Description

+ Stored property 'model' of 'Sendable'-conforming struct 'LinesShape' has non-sendable type 'ChartModel'

+ Stored property 'chartType' of 'Sendable'-conforming struct 'ColumnChartCategoryShape' has non-sendable type 'ChartType'

+ Stored property 'chartContext' of 'Sendable'-conforming struct 'PointsShape' has non-sendable type 'any ChartContext'

### DIFF
--- a/Sources/FioriCharts/Charts/Common/ChartContext.swift
+++ b/Sources/FioriCharts/Charts/Common/ChartContext.swift
@@ -2,7 +2,7 @@ import Combine
 import Foundation
 import SwiftUI
 
-protocol ChartContext: AnyObject {
+protocol ChartContext: AnyObject, Sendable {
     func scaleX(_ model: ChartModel, plotViewSize: CGSize) -> CGFloat
     
     func scaleY(_ model: ChartModel, plotViewSize: CGSize) -> CGFloat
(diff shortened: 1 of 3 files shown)


billzhou0223 (approved):

Sources/FioriCharts/Model/ChartModel.swift
```diff
@@ -1855,6 +1855,8 @@ extension ChartModel: CustomStringConvertible {
     }
 }
 
+extension ChartModel: @unchecked Sendable {}
```
dyongxu, line 1858: Not safe to use @unchecked 

e.g.  in ChartModel.swift , line 101
   @Published public var chartType: ChartType 
 Compiler will skip to check this line for "Sendable" conformance. However, since chartType is "var", it is not thread safe.

Sources/FioriCharts/Charts/Common/ChartContext.swift
```diff
@@ -42,6 +42,8 @@ protocol ChartContext: AnyObject {
     func closestSelectedPlotItems(_ model: ChartModel, atPoints: [CGPoint], rect: CGRect, layoutDirection: LayoutDirection) -> [(Int, Int)]
 }
 
+extension DefaultChartContext: @unchecked Sendable {}
```
dyongxu, line 45: Use of @unchecked is not safe.

dyongxu (changes requested):
Remove the "@unchecked" change.
Leave the warning message there for now for these two files.
We may have to do some re-code for these two files to be thread safe.

KevinZK:
Due to thread safety issues, the warning is temporarily retained and PR is closed.

PR #715, opened 2024-06-13
chore: 🤖[HCPSDKFIORIUIKIT-2608] Clean up warnings: Logical problem

KevinZK:
# Description

+ Comparing non-optional value of type 'Bool' to 'nil' always returns true


dyongxu (approved):

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
```diff
@@ -266,7 +266,7 @@ extension SignatureCaptureView: View {
         if isRequired {
             let requiredText = NSLocalizedString("Required Field", tableName: "FioriSwiftUICore", bundle: Bundle.accessor, comment: "Required Field")
             let labelString = _title?.suffix(1) == "*" ? String(_title?.dropLast(1) ?? "") : (_title ?? "")
-            accString = labelString + (labelString.isEmpty != nil ? ", " : "") + requiredText
+            accString = labelString + (labelString.isEmpty ? ", " : "") + requiredText
```
billzhou0223, line 269: Shouldn't the logic be `labelString.isEmpty ? "" : ", "`

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/SignatureCaptureView+View.swift
KevinZK: fixed

billzhou0223 (approved):

PR #723, opened 2024-06-24
feat: 🎸 [JIRA:IOSSDKBUG-5] add variables to ChartLabelAttributes

hengyi-zhang:
add two variables(font and backgroundColor) to ChartLabelAttributes, for user to custom chart label attributes conveniently

✅ Closes: #692


dyongxu (approved):
Approved as a cherrypick from main

PR #728, opened 2024-06-27
chore: 🤖 fix fatalError when running tests of a package that relies on FioriSwiftUI package

MarcoEidinger:
Inspiration from https://forums.swift.org/t/unable-to-find-bundle-in-package-target-tests-when-package-depends-on-another-package-containing-resources-accessed-via-bundle-module/43974/5


MarcoEidinger:
fyi: discovered when trying to run `ViewInspector` based unit tests for JMC

billzhou0223 (approved):

PR #729, opened 2024-06-28
fix: 🐛 ignore small diff for frame reader

xiaoyu0722:


Sources/FioriSwiftUICore/Experimental/Popover/View+Extension.swift
```diff
@@ -47,3 +27,54 @@ struct ContentSizeReaderPreferenceKey: PreferenceKey {
     static var defaultValue: CGSize { .zero }
     static func reduce(value: inout CGSize, nextValue: () -> CGSize) { value = nextValue() }
 }
+
+struct FrameReaderModifier: ViewModifier {
```
billzhou0223, line 31: May I ask why do you need to reply on `ContentFrameReaderPreferenceKey` rather than just calling `rect(frame)` in GeometryReader. 

You are not setting `lastFrame` anywhere.

Sources/FioriSwiftUICore/Experimental/Popover/View+Extension.swift
```diff
@@ -47,3 +27,54 @@ struct ContentSizeReaderPreferenceKey: PreferenceKey {
...
+                                self.rect(newValue)
+                            }
+                        }
+                }
+                .hidden()
+            )
+    }
+}
+
+struct SizeReaderModifier: ViewModifier {
```
billzhou0223, line 58: Same question here.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Experimental/Popover/View+Extension.swift
xiaoyu0722: Sorry, I missed the `self.lastFrame = newValue`
And the preference key is to get value changed constantly, and throw the value by the last closure.

billzhou0223: Preference is useful for propagating value to its parent view. In your case, the sender and the receiver is the same view so there is no need to use preference I think.

```
func body(content: Content) -> some View {
        content
            .background(
                GeometryReader { geometry in
                    let frame = geometry.frame(in: self.coordinateSpace)
                    if abs(frame.width - self.lastFrame.width) > 0.001 ||
                                abs(frame.height - self.lastFrame.height) > 0.001 ||
                                abs(frame.origin.x - self.lastFrame.origin.x) > 0.001 ||
                                abs(frame.origin.y - self.lastFrame.origin.y) > 0.001
                            {
                                self.lastFrame = frame
                                self.rect(newValue)
                            }

                    Color.clear
                }
            )
    }
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Experimental/Popover/View+Extension.swift
xiaoyu0722: Can not call `rect(newValue)` directly in geometry reader.

billzhou0223: Why?

billzhou0223: What's the error you see if you call it in GeometryReader?

xiaoyu0722:
I will close this PR and no need to update this.

PR #735, opened 2024-07-08
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2631] RatingControlFormView SwiftUI implementation

janhuachu:
[JIRA:HCPSDKFIORIUIKIT-2631] RatingControlFormView SwiftUI implementation

✅ Closes: JIRA:HCPSDKFIORIUIKIT-2631


Sources/FioriSwiftUICore/_FioriStyles/RatingControlFormViewStyle.fiori.swift
```diff
@@ -0,0 +1,159 @@
+import FioriThemeManager
+import Foundation
+import SwiftUI
+
+// Base Layout style
+public struct RatingControlFormViewBaseStyle: RatingControlFormViewStyle {
+    public func makeBody(_ configuration: RatingControlFormViewConfiguration) -> some View {
+        // Add default layout here
+        VStack {
+            if !configuration.subtitle.isEmpty {
```
JunSong-SH, line 10: Nor sure what the logic is based on or what the requirements are. However, we might consider displaying the subtitle, when it is not empty, either in one column or alongside the Title and RatingControl in one row. Then, the subtitle could be utilized as the value label, as described in JIRA [FIORIDESIGN-11992](https://jira.tools.sap/browse/FIORIDESIGN-11992). Consequently, we wouldn't need to add another 'Text' element for the value label. What do you think about this approach? If you agree, you can go ahead and make the changes, or I can modify it later.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -317,3 +317,11 @@ protocol _RatingControlComponent {
     /// The custom spacing between images.
     var interItemSpacing: CGFloat? { get }
 }
+
+/// The form view which contains a title, rating control, and a subtitle
+// sourcery: CompositeComponent
+protocol _RatingControlFormViewComponent: _TitleComponent, _RatingControlComponent, _SubtitleComponent, _FormViewComponent {
+    /// Indicates if the title and rating control is on a single horizontal row or not.
+    // sourcery: defaultValue = "false"
+    var isStacked: Bool { get }
```
billzhou0223, line 326: Please name it `var axes: Axis`. (See `ScrollView`)

Sources/FioriSwiftUICore/_FioriStyles/RatingControlFormViewStyle.fiori.swift
```diff
@@ -0,0 +1,159 @@
...
+                .titleStyle { titleConf in
+                    Title(titleConf)
+                        .foregroundStyle(self.getTitleColor(configuration))
+                        .font(.fiori(forTextStyle: .subheadline, weight: .semibold))
+                }
+                .subtitleStyle { subtitleConf in
+                    Subtitle(subtitleConf)
+                        .foregroundColor(self.getSubtitleColor(configuration))
+                        .font(.fiori(forTextStyle: .footnote))
+                }
```
billzhou0223, line 106: Title style and subtitle style should go into their corresponding style definition.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/RatingControlFormViewStyle.fiori.swift
janhuachu: I think the subtitle is something that the developer to put for describing the rating is for.
It is different than what [FIORIDESIGN-11992](https://jira.tools.sap/browse/FIORIDESIGN-11992) is trying to add, which is a value description. I think we should not reuse the subtitle field.

Sources/FioriSwiftUICore/_FioriStyles/RatingControlFormViewStyle.fiori.swift
janhuachu: OK

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
janhuachu: OK

janhuachu:
@billzhou0223 Updated per your comments. Please review again.

billzhou0223 (approved):

Sources/FioriSwiftUICore/_FioriStyles/RatingControlFormViewStyle.fiori.swift
JunSong-SH: OK. In that case, we don't consider to reuse it.

JunSong-SH: OK. In that case, we don't consider to reuse it.

JunSong-SH (approved):

Sources/FioriSwiftUICore/_generated/StyleableComponents/RatingControlFormView/RatingControlFormView.generated.swift
```diff
@@ -0,0 +1,143 @@
+// Generated using Sourcery 2.1.7 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+import Foundation
+import SwiftUI
+
+/// The form view which contains a title, rating control, and a subtitle
+public struct RatingControlFormView {
+    let title: any View
+    /// The rating value.
+    @Binding var rating: Int
```
JunSong-SH, line 10: BTY, not sure if you can change the rating to float as we need support half star later. Also, in RatingControl. We might need App team to pass in rating value like 2.1, 3.7 etc,. I ask this since I am not sure if I can finish the later change to support half star in 24.8. If not, we need change the API in 24.12.

PR #748, opened 2024-07-23
fix: 🐛 [JIRA:IOSSDKBUG-264] No left padding for the 1st column in DataTable

restaurantt:
fix item padding in regular or compact mode, also in edit mode


billzhou0223:
Need a review from @shengxu7

billzhou0223 (approved):

PR #753, opened 2024-07-30
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2686] Button update

hengyi-zhang:


Sources/FioriSwiftUICore/FioriButton/FioriButton.swift
```diff
@@ -205,3 +206,110 @@ private struct _ButtonStyleImpl: ButtonStyle {
...
+    /// Create a fiori button with loading indicator (or success icon) and title text.
+    /// - Parameters:
+    ///   - isSelectionPersistent: A boolean value determines whether the button should be persistent or not.
+    ///   - action: Action triggered when tap on button.
+    ///   - title: A closure that returns a title text for each state. For a button with non-persistent selection, `.normal`, `.disabled`, `.highlighted` are supported. For a button with persistent selection, use `.selected` instead of `.highlighted`.
+    ///   - image: image of the button when loading state is .unspecified.
+    ///   - loadingState: A state to show activity indicator or success icon, default is .unspecified.
+    ///   - imagePosition: Place the image along the top, leading, bottom, or trailing edge of the button.
+    ///   - imageTitleSpacing: spacing between image and title.
+    init(isSelectionPersistent: Bool = false,
```
billzhou0223, line 243: Let's extend existing init to support image display.

```
public init(isSelectionPersistent: Bool = false,
                action: ((UIControl.State) -> Void)? = nil,
                @ViewBuilder label: @escaping (UIControl.State) -> any View，
                @ViewBuilder image: @escaping (UIControl.State) -> any View = { _ in },
                imagePosition: Position = .left,
                spacing: CGFloat = 8)
```
You need to update other init accordingly as well. `FioriButtonStyleConfiguration` needs to be updated as well.

We need to provide default UI based on loading state. We can add a new style to achieve that.

```
FioriButton()
    .fioriButtonStyle(FioriLoadingButtonStyle(state: .inProgress // or .success))
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/FioriButton/FioriButton.swift
hengyi-zhang: Fixed!

billzhou0223 (approved):
LGTM!

PR #754, opened 2024-07-31
fix: fix compilation error caused by StepperView

billzhou0223:


dyongxu (changes requested):
Need to fix error  in other files as well. These errors will show if compiling the example app.
scrollDismissesKeyboard not available on visionOS

MarcoEidinger (commented):
Verified framework compilation locally by following steps mentioned in https://github.com/SAP/cloud-sdk-ios-fiori/blob/main/.xcodegen/README.md

Screenshot 2024-07-31 at 12 04 46 PM

dyongxu (approved):

PR #757, opened 2024-08-01
fix: IOSSDKBUG-282 - JMC welcome screen legal disclaimer bg color

CharlesXu0488:
Adding default JMC welcome screen background color so that the legal disclaimer's background color matches conversation scroll view's background color.

After adding background color
image

Before adding background color
image

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/JouleWelcomeScreenStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/JouleWelcomeScreenStyle.fiori.swift
@@ -38,7 +38,6 @@ public struct JouleWelcomeScreenBaseStyle: JouleWelcomeScreenStyle {
             Spacer()
             configuration.footnote
         }
-        .background(Color.preferredColor(.primaryBackground))
     }
 }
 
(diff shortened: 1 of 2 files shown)


dyongxu (approved):

billzhou0223 (approved):

Sources/FioriSwiftUICore/_FioriStyles/JouleWelcomeScreenStyle.fiori.swift
```diff
@@ -38,6 +38,7 @@ public struct JouleWelcomeScreenBaseStyle: JouleWelcomeScreenStyle {
             Spacer()
             configuration.footnote
         }
+        .background(Color.preferredColor(.primaryBackground))
```
billzhou0223, line 41: You should add the background in the `JouleWelcomeScreenFioriStyle.ContentFioriStyle`. The base style is only for layout purpose. Styling attributes should be set in corresponding fiori style.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/JouleWelcomeScreenStyle.fiori.swift
CharlesXu0488: OK, done.

billzhou0223 (approved):

dyongxu (approved):

PR #772, opened 2024-08-13
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2702]Illustrated Message Enhancement

xiaoqinggrace:


billzhou0223:
@JuliaSchnSAP Can you help do a review given your experiences for this component?

Sources/FioriSwiftUICore/_generated/StyleableComponents/IllustratedMessage/IllustratedMessage.generated.swift
```diff
@@ -8,7 +8,13 @@ public struct IllustratedMessage {
     let title: any View
     let description: any View
     let action: any View
+    let secondaryAction: any View
+    /// Specifies the size of the detailImage. This value should be an enum from the DetailImageSize set.
     let detailImageSize: IllustratedMessage.DetailImageSize?
+    /// Determines the layout of the action buttons. If set to true, the buttons will be arranged vertically. If set to false, they will be arranged horizontally. The default value is true.
+    let isActionVerticalAligned: Bool
```
JuliaSchnSAP, line 15: I would call this parameter "isActionVerticallyAligned"

JuliaSchnSAP:
In the mixed layout without setting any of the other parameters, the text is center-aligned if only a title is shown but it is left-aligned if there is also a description

image image

JuliaSchnSAP (approved):

xiaoqinggrace:
@JuliaSchnSAP In the two use cases of mixed mode shown in the screenshot, the title and description are all leading alignment. It is the the default setting. I could not find the center alignment case in this setting.

JuliaSchnSAP:
When the layout axis is unset, the button is full width by default, but when it is set to a vertical layout the button has a fixed size by default. Was this part of the design spec?

Old version of code - unset layout axis
image 

PR - unset layout axis
image

PR - vertical layout
image

JuliaSchnSAP:
> @JuliaSchnSAP In the two use cases of mixed mode shown in the screenshot, the title and description are all leading alignment. It is the the default setting. I could not find the center alignment case in this setting.

I think the text is being centered even though you didn't intend to do so. I actually noticed the description also has a center alignment if the view's width is large enough. It might be because the outermost VStack doesn't have an alignment set so it defaults to center alignment. Here is another screenshot of the mixed layout with a larger width:

image

xiaoqinggrace:
As for the vertical style case, the fixed width button is by default when the .vertical style applied. It is the new update in the current spec. 
As for the mix style case, the leading alignment is by default for the title and description in its container. In most cases, the image will be shown on the left, the title and description are leading aligned. The illustrated message view is central aligned in the container. The last screenshot is the special case for the mix style.

Sources/FioriSwiftUICore/_FioriStyles/IllustratedMessageStyle.fiori.swift
```diff
@@ -29,11 +20,30 @@ public struct IllustratedMessageBaseStyle: IllustratedMessageStyle {
...
+            } else {
+                HStack {
+                    configuration.secondaryAction
+                        .fioriButtonStyle(FioriSecondaryButtonStyle(colorStyle: .tint, maxWidth: .infinity))
+                    configuration.action
+                        .fioriButtonStyle(FioriPrimaryButtonStyle()) // .infinity))
+                }
+                .padding(.top, 16)
+            }
         }
```
JuliaSchnSAP, line 45: In all of the layouts, the 16pt padding above the buttons is shown even when the buttons are not being rendered

image

dyongxu (approved):

PR #784, opened 2024-08-28
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2716]Step Progress Indicator Enhancement

KevinZK:
* Add image support for the Step Progress Indicator Component

Simulator Screenshot - iPhone 15 Pro - 2024-08-28 at 17 49 50

### DIFF
--- a/Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
+++ b/Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
@@ -98,8 +98,12 @@ struct DefaultSingleStep: View {
             } node: {
                 ZStack {
                     self.node(by: self.stepItem.state, isSelected: isSelected)
-                    Text("\(self.index + 1)")
-                        .font(Font.fiori(forTextStyle: .footnote))
+                    if self.stepItem.icon.isEmpty {
+                        Text("\(self.index + 1)")
+                            .font(Font.fiori(forTextStyle: .footnote))
+                    } else {
+                        self.stepItem.icon
+                    }
                 }
                 .frame(width: self.sideLength, height: self.sideLength)
                 .overlay {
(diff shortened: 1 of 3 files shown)


Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
```diff
@@ -98,8 +98,12 @@ struct DefaultSingleStep: View {
             } node: {
                 ZStack {
                     self.node(by: self.stepItem.state, isSelected: isSelected)
-                    Text("\(self.index + 1)")
-                        .font(Font.fiori(forTextStyle: .footnote))
+                    if self.stepItem.icon.isEmpty {
```
billzhou0223, line 101: Have you tested this code? I suspect that `Image.isEmpty` won't return the correct value.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
KevinZK: The `isEmpty` extension was newly added by you on February 2, and the result of my test is valid.

Implementation
```
public protocol _ViewEmptyChecking {
    // Check if the view is an empty view. Some SDK components implement this protocol to define their own empty logic.
    var isEmpty: Bool { get }
}
extension View {
    var isEmpty: Bool {
        if self is EmptyView {
            return true
        }
        
        if let self = self as? _ViewEmptyChecking {
            return self.isEmpty
        }
        
        return false
    }
}
extension Optional: _ViewEmptyChecking where Wrapped: View {
    public var isEmpty: Bool {
        if case .none = self {
            return true
        }
        
        if case .some(let wrapped) = self {
            return wrapped.isEmpty
        }
        
        return false
    }
}
```

billzhou0223 (approved):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
```diff
@@ -98,8 +98,12 @@ struct DefaultSingleStep: View {
...
+                        Text("\(self.index + 1)")
+                            .font(Font.fiori(forTextStyle: .footnote))
+                    } else {
+                        self.stepItem.icon
+                    }
```
xiaoyu0722, line 106: `Icon steps and numeric steps cannot be used together.`
Icons and number text may mix together by this logic. Please double check with designer.

PR #787, opened 2024-09-02
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2716]Step Progress Indicator

KevinZK:
~~*Add stepProgressIndicatorNodeType for step node.~~

## Description
~~Add `StepProgressIndicatorNodeType` for stepProgressIndicator to set the nodetype style.~~

~~+ mixture: Allow nodes to display numeric step and icon step in a mixture.~~
~~+ icon: Only allow the display of icon step.~~
~~+ text：Only allow the display of numeric step.~~

add `TextorIcon` replace `Image`


dyongxu:
@xiaoyu0722 Can you review this enhancement as well ?

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
```diff
@@ -14,3 +14,13 @@ public protocol StepItem {
     /// Node icon.
     var icon: Image? { get }
 }
+
+/// Node of Step Progress Indicator display type, default is `mixture`.
+public enum StepProgressIndicatorNodeType {
```
billzhou0223, line 19: There is already a type for this purpose called `TextOrIcon`. The node should be declared as this type so that developers can pass either text or icon as needed.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
```diff
@@ -11,6 +11,16 @@ public protocol StepItem {
...
-    /// Node icon.
-    var icon: Image? { get }
+    /// Step node.
+    var node: TextOrIcon? { get }
+}
+
+/// Node of Step Progress Indicator display type, default is `mixture`.
+public enum StepProgressIndicatorNodeType {
+    /// Text and icon.
+    case mixture
```
xiaoyu0722, line 21: If this is not supported, why public it?
Can we handle this by a default empty icon(like dashed box in Figma) or just ignore icons and use index directly by data checking?  Add a statement in doc that how we handle the mixed data.

xiaoyu0722 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
KevinZK: Only an option is provided, allowing developers to decide for themselves. The placeholder icon 'app.dashed' has been provided.

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
```diff
@@ -11,6 +11,16 @@ public protocol StepItem {
     var state: StepProgressIndicatorState { get set }
     /// Substeps for this one.
     var substeps: [StepItem] { get set }
-    /// Node icon.
-    var icon: Image? { get }
+    /// Step node.
+    var node: TextOrIcon? { get }
```
billzhou0223, line 15: This is a breaking change. Use available tag in such case.

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
billzhou0223: You don't need this type after adopting `TextOrIcon`.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
KevinZK: `var icon: Image? { get } `was added in the previous PR and should not be a breaking change.

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
KevinZK: Sorry, I don't understand why. 

My thinking is like this.
`TextOrIcon` is only applicable to Node. The enumeration `StepProgressIndicatorNodeType` is the display form of all nodes in the StepProgressIndicator component.

Anyway, my latest commit has removed `mixture`.

billzhou0223: I mean we can remove this enum type `StepProgressIndicatorNodeType` not just one of the cases. In the old implementation made by @xiaoyu0722, if the icon is nil we show text by default. Otherwise we use the custom icon from user. I'd like to keep this behavior. 

Where did you get this requirement of allowing switching between text and icon for the default node? If it's not something on the spec, don't change it.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
billzhou0223: A breaking change means you changed or removed a public api that was already released... It will break the app code that relies on it. We need to use available keyword for such changes.

BTW, do not mark a problem as resolved if you are not sure what you did is correct or not. Otherwise I might miss these issues in the next review.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
KevinZK: Designer's requirements“Icon steps and numeric steps cannot be used together”

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
KevinZK: I mean that this PR and #784 are actually both modifications to the Step Progress Indicator. The modification of var icon: Image? { get } is newly added in #784.
Is it still necessary to continue marking "available" in this situation?

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
billzhou0223: That's more like a guideline to app developers. We don't want to add this restriction to this component.

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepItem.swift
billzhou0223: No if you recently added it and hasn't been released

KevinZK: Got it

Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
```diff
@@ -100,23 +99,19 @@ struct DefaultSingleStep: View {
             } node: {
                 ZStack {
                     self.node(by: self.stepItem.state, isSelected: isSelected)
-                    self.noNodeStep()
-                    switch self.type {
-                    case .icon:
+                    if self.stepItem.node != nil {
```
billzhou0223, line 102: You can use `TextOrIconView` here.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/_DefaultSteps.swift
KevinZK: `TextOrIconView` makes code more concise.

KevinZK:
CI Error：

The actions/download-artifact in CI.yml is version v1.
Is it updated to actions/download-artifact@v2?

billzhou0223 (approved):

billzhou0223:
> CI Error：
> 
> The actions/download-artifact in CI.yml is version v1. Is it updated to actions/download-artifact

Version was taken care of by #799

xiaoyu0722 (approved):

PR #788, opened 2024-09-03
feat: 🎸 [JIRA: HCPSDKFIORIUIKIT-2708] avatar stack support

xiaoyu0722:
|grouped|horizontal|
|-|-|
|Simulator Screenshot - iPhone 15 Pro Max - 2024-09-03 at 17 04 52|Simulator Screenshot - iPhone 15 Pro Max - 2024-09-03 at 17 05 00|


Apps/Examples/Examples/FioriSwiftUICore/AvatarStack/AvatarStackExample.swift
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+        .avatarsLayout(self.model.avatarsLayout)
+        .isAvatarCircular(self.model.isCircular)
+        .avatarsTitlePosition(self.model.titlePosition)
+        .avatarsSpacing(self.model.spacing)
+        .avatarsMaxCount(self.model.maxCount)
+        .avatarBorderColor(self.model.borderColor)
+        .avatarBorderWidth(self.model.borderWidth)
+        .avatarSize(self.avatarSize)
```
billzhou0223, line 29: Environment is useful when you have multiple UI elements of the same type in the hierarchy and you want to be able to set something in the parent and propagate down to all the children. In your case, I don't see this type of requirement.

Let's try to make some of the properties as init params. The others can be configured through style modifier.

```
AvatarStack(layout: .grouped, spacing: 8, ...)
      .avatarsStyle {
             Avatars($0)
                  .borderColor()
                  .borderWidth()
      }
```

billzhou0223 (changes requested):

Apps/Examples/Examples/FioriSwiftUICore/AvatarStack/AvatarStackExample.swift
xiaoyu0722: These environments are for uniform settings of avatars.
`.avatarsStyle {}` is for the whole `avatars` which is new grouped view generated by `AvatarBuilder`.

If developer want to set different styles for different avatars, they should do the styles setting in the avatars initialization like below way.

Screenshot 2024-09-05 at 14 39 45

billzhou0223: OK let's keep them as environment values. But at lease we can combine border width and color into one api. Same as SwiftUI's border modifier.

xiaoyu0722: Updated to `func avatarsBorder(_ color: Color, width: CGFloat = 1) -> some View`

billzhou0223 (approved):

PR #790, opened 2024-09-03
fix: 🐛 [JIRA:IOSSDKBUG-324]Usage of SwiftUI KeyValueFormView

xiaoqinggrace:
✅ Closes: IOSSDKBUG-324


dyongxu (changes requested):
Add a test case to show this NoteForm as a standalone view. i.e. it is not included in a list.

dyongxu (approved):

billzhou0223 (approved):

PR #794, opened 2024-09-05
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2707] SwiftUI RatingControl Enhancement

janhuachu:
SwiftUI RatingControl enhancement with value label and review count label

✅ Closes: HCPSDKFIORIUIKIT-2707


Sources/FioriSwiftUICore/DataTypes/RatingControl+DataType.swift
```diff
@@ -1,38 +1,53 @@
...
+         This `RatingControl` is read-only. Each rating star is in small scale.
          */
         case standard
 
+        /**
+         Standard large style.
+
+         This `RatingControl` is read-only. Each rating star is in large scale.
+         */
+        case standardLarge
```
JunSong-SH, line 36: I'd like to use the `Bool` variable `isLarge` with the UIKit `FUIRatingControl`. By default, it is set to `false` for the read-only state and always `true` for the editable and disabled states. This approach is more simple to me when a smaller size is needed for the editable or disabled states, or for any new states that may be introduced in the future. For us, this means that we do not need to add too many styles, and for the App team, it makes it easy to understand that there are two sizes for each state. 
I just want to let you know about the difference between your API and mine. However, I am definitely okay with your implementation for these two new styles here for SWiftUI Rating Control.

Sources/FioriSwiftUICore/_FioriStyles/RatingControlStyle.fiori.swift
```diff
@@ -6,13 +6,37 @@ import SwiftUI
 public struct RatingControlBaseStyle: RatingControlStyle {
     public func makeBody(_ configuration: RatingControlConfiguration) -> some View {
         HStack(spacing: configuration.getItemSpacing()) {
-            ForEach(configuration.ratingItems(configuration.rating)) { ratingItem in
-                if ratingItem.isOn {
-                    configuration.getOnImageView()
-                } else {
-                    configuration.getOffImageView()
+            if configuration.showsValueLabel, self.getReadOnly(configuration) {
```
JunSong-SH, line 9: I am a little confused about the left value for the editable/disabled state. The design spec mentions 'Left value (optional) except for editable and disabled variants.'. From my understanding, this means that the left value is required/mandatory for the editable and disabled states. I have consulted the designer about this in Figma. 
There's a high probability that my understanding of the above sentence might be incorrect. But, just want to let you know this. :-)

Sources/FioriSwiftUICore/_FioriStyles/RatingControlStyle.fiori.swift
```diff
@@ -6,13 +6,37 @@ import SwiftUI
...
+            } else {
+                ForEach(configuration.ratingItems(configuration.rating)) { ratingItem in
+                    if ratingItem.isOn {
+                        configuration.getOnImageView()
+                    } else {
+                        configuration.getOffImageView()
+                    }
                 }
             }
+            if configuration.showsValueLabel, !self.getReadOnly(configuration) {
```
JunSong-SH, line 32: Now, I understand the meaning of valueLabel and reviewCountLabel you always mentioned during our discussion But, for me, I would think it is the value concept for displaying on left and right. And, I feel it is not very straightforward. For example, here, if App team want to customize the value displayed on the right, they must to know when they should set the value for one of 'ratingValueFormat', 'averageRatingFormat', 'reviewCountFormat' and 'reviewCountCeilingFormat'. 

For me, I'd like to use more simple or straightforward logic on UIKit: To use two 'showXXXLabel' bool flag to control label display and use two 'XXXLabelFormat' for customization whatever we call it 'leftLabel/leadingLabel/valueLabel' or 'rightLabel/trailingLabel/reviewCountLabel'(Now, I use valueLabel and reviewCountLabel base on our discussion before). Absolutely, 'reviewCount' and 'reviewCountCeiling' were needed. Then, App team just to know there are two 'XXXLabelFormat' for customization and the default value for those two labels was that the label on left will use default formatted rating value or average rating value according to the style(confirming with designer about should display value for NON-readonly state). For the value of label on right, by default, to use '3 of 5' or '2.3 of 5' if there is no review count and use '9 review' or '8+ reviews' if there has reviewCount or reviewCount>reviewCountCeiling.

Certainly, I just want to let you know about the different between your API and mine. :-)

JunSong-SH (approved):
LGTM!

Sources/FioriSwiftUICore/DataTypes/RatingControl+DataType.swift
```diff
@@ -53,7 +69,25 @@ extension RatingControlConfiguration {
...
+    func ratingItems(_ averageRating: CGFloat) -> [RatingItem] {
+        var items: [RatingItem] = []
+        for i in self.ratingBounds {
+            guard i != self.ratingBounds.upperBound else {
+                continue
+            }
+            let diff = averageRating - CGFloat(i)
+            if diff < 0.3 {
+                items.append(RatingItem(isOn: false, isHalf: false))
+            } else if diff < 0.7 {
```
JunSong-SH, line 86: See FIORITECHE1-8222, web component use 1.0-1.2 -> 1, 1.3-1.7 -> 1.5 and 1.8-1.9 -> 2. So, here, shouldn't diff <= 0.7 or diff < 0.8?

Sources/FioriSwiftUICore/DataTypes/RatingControl+DataType.swift
janhuachu: @JunSong-SH I was originally using the isLarge property. However, later I removed it to reduce the number of properties. I think there are too many properties already.

Sources/FioriSwiftUICore/_FioriStyles/RatingControlStyle.fiori.swift
janhuachu: @JunSong-SH I have not thought about the behavior you mentioned above. Let's wait for the designer to clarify the behavior.

Sources/FioriSwiftUICore/DataTypes/RatingControl+DataType.swift
janhuachu: Yes, you're right. I'll change it.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -314,6 +314,9 @@ protocol _RatingControlComponent {
     /// The custom image to be used for "Off".
     var offImage: Image? { get }
 
+    /// The custom image to be used for half On image.
+    var halfImage: Image? { get }
```
billzhou0223, line 318: Why are these `Image` properties not view builder? Like title, subtitles, those UI elements should be declared as view builder so that they become customizable.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -325,6 +328,51 @@ protocol _RatingControlComponent {
...
+    // sourcery: defaultValue = "nil"
+    var ratingValueFormat: String? { get }
+
+    /// The custom font for value label.
+    // sourcery: defaultValue = "nil"
+    var valueLabelFont: Font? { get }
+
+    /// The custom text color for value label.
+    // sourcery: defaultValue = "nil"
+    var valueLabelColor: Color? { get }
```
billzhou0223, line 343: I don't think these style props are needed, because we can customize them using style modifiers

```
RatingControl()
    .valueLabelStyle {
        ValueLabel($0)
            .font()
            .foregroundStyle()
    }
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
janhuachu: OK. I'll change that.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
janhuachu: OK. I'll change it. Thanks.

billzhou0223 (approved):

PR #809, opened 2024-10-02
feat: 🎸 HCPSDKFIORIUIKIT-2718 Mandatory Field for DateTimePicker

xiaoqinggrace:
Add Mandatory Field Styling for DateTimePicker. Add controlState for disable/readonly state. Add informationView for the error/hint message.

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/Picker/DateTimePickerExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/Picker/DateTimePickerExample.swift
@@ -3,14 +3,25 @@ import Foundation
 import SwiftUI
 
 struct DateTimePickerExample: View {
-    @State var s1: Date = .init()
+    @State var s1: Date = .init(timeIntervalSince1970: 0.0)
     @State var s2: Date = .init()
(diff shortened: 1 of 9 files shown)


Apps/Examples/Examples/FioriSwiftUICore/Picker/DateTimePickerExample.swift
```diff
@@ -3,10 +3,13 @@ import Foundation
 import SwiftUI
 
 struct DateTimePickerExample: View {
-    @State var s1: Date = .init()
+    @State var s1: Date = .init(timeIntervalSince1970: 0.0)
```
angiexyang, line 6: Maybe we can make selectedDate an optional value for developers, and set  `Date(timeIntervalSince1970: 0.0)` internally if selectedDate is empty. In UIKit, developers can set `cell.value = nil` to show "No Date Selected"

xiaoqinggrace: If we implement this, we need to change the model and ensure the application passes the binding value.

angiexyang (approved):

PR #823, opened 2024-10-14
feat: 🎸 [JIRA: HCPSDKFIORIUIKIT-2764] list picker enhancement

xiaoyu0722:
| Debug Panel | Normal Single Selection | Multiple Selections |
| --- | --- | --- |
|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 00 09|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 25 36|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 43 59|

| Search | Filter | Cancel Alert | Customizaiton |
| --- | --- | --- | --- |
|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 44 19|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 44 29|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 45 39|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-14 at 17 46 12|


Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -569,3 +569,37 @@ protocol _DateTimePickerComponent: _TitleComponent, _ValueLabelComponent, _Manda
 
 // sourcery: CompositeComponent
 protocol _AvatarStackComponent: _AvatarsComponent, _AvatarsTitleComponent {}
+
+/// `ListPickerItem` is a view that provide a `NavigationLink` with a title and selected value. And `ListPickerDestination` is recommended to be used as its destination, which selection, search filter and customized rows are supported.
```
dyongxu, line 573:
```suggestion
/// `ListPickerItem` is a view that provides a `NavigationLink` with a title and selected value(s). And `ListPickerDestination` is recommended to be used as its destination, for which selection, search filter and customized rows are supported.
```

Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
```diff
@@ -0,0 +1,185 @@
+import FioriSwiftUICore
+import SwiftUI
+
+struct ListPickerItemExample: View {
```
dyongxu, line 4: In UIKit, the list picker can group some values in sections. 
e.g.
Section 1 for all apples
Section 2 for all pears
Section 3 for all grapes

Currently, looks like only "Selected" and "All" section are supported. 
Do this new implementation support multiple "group" idea as well ? If yes, can we add some new cases?

dyongxu (changes requested):
comments inlined

Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
xiaoyu0722: Currently, not support different groups, I will check again and see how to make it works.

Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
```diff
@@ -0,0 +1,760 @@
...
+        )
+        self.init {
+            FioriButton { _ in Text("Cancel".localizedFioriString()) }
+        } listPickerContent: {
+            content
+        }
+    }
+}
+
+public struct ListPickerDestinationContent<Data: RandomAccessCollection, ID: Hashable, RowContent: View>: View {
```
billzhou0223, line 296: Do developers need to have access to this type? If not we can make it internal.

Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
```diff
@@ -0,0 +1,189 @@
...
+    @State var disableEntriesSection: Bool = false
+    @State var allowEmpty: Bool = false
+    @State var autoDismissDestination: Bool = false
+
+    var body: some View {
+        List {
+            ListPickerItem(title: {
+                Text("Frameworks")
+            }, value: {
+                self.valueView
```
billzhou0223, line 29: In your example, developers need to observe selection changes and construct the value view by themselves. This is good because it's very customizable. In addition to that, do you also provide a default implementation for building the value view in SDK so that this component becomes more convenient to use? For example, the default value view in the simplest form could be concatenating the ids of selected items.

Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
xiaoyu0722: No, I will remove public statement.

Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
xiaoyu0722: Emm, api takes id as an identifier. I'm not sure if it's a good idea to take it to generate a default value view.
The example use a separate `valueView` is for different cases, like optional/empty/multiple values.

billzhou0223: Maybe using id is not good. How about introduce a new param `name: KeyPath<Data.Element, String>?` in `ListPickerDestination`'s init which can be used as default selection name. 

I think we need to provide a default value view according to spec.

dyongxu (approved):

dyongxu:
@xiaoyu0722 Need to resolve merge conflict issue.

Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
xiaoyu0722: Hi @billzhou0223 Any other issue? I will merge first if no other problem.

billzhou0223: If we don't have a default implementation, how can you ensure the component's value field matches the spec? Having the developers to create the value view on their own and set the required UI properties like font, color, accessibility attributes, etc according to spec is not realistic and it deviates the purpose of providing it as a SDK component.

What you need is a way to communicate between different components and I can think of several on the top of my head like callback, pub-sub (Combine), notification (not recommended), etc.

xiaoyu0722: Is this not the purpose of styleable template that allow customization?
If they just a clean text used for value component, then fiori style will be applied. If not, then it's up to developer?

billzhou0223: Of course customization is important and that's well supported currently with the new architecture. But in addition to that, SDK component should provide convenience and allow developers to achieve more by doing less. Based on the common use case of this component, it's easy to understand that the value is always constructed from the selected items so why don't we provide a default implementation to help developers take care of this tedious work? It could be simply a text as you said but it could also be something else more complicated (designers wanted to replace them with [chips](https://www.figma.com/design/OKmSSho6WyEL7c4sWiIqXc/List-Picker-Form-Cell-(iOS)?node-id=1-18&node-type=canvas&t=bZ6DLLgGjGuEMRvh-0) before but we didn't implement it). Having this default implementation in place allows us to adapt to future design changes more easily. Hope this makes sense.

xiaoyu0722: Hi @billzhou0223 
A default `Text` is added when value is empty in last commit.
But there is a problem, that developer can not use `.valueStyle {}` to modify its style when they do not give a value view.

billzhou0223: I think you need to wrap the default view with a `Value` view.

xiaoyu0722: Fixed.

billzhou0223 (approved):

PR #826, opened 2024-10-16

hengyi-zhang:
Title and icon support developer customization in banner message

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/BannerMessageStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/BannerMessageStyle.fiori.swift
@@ -20,7 +20,7 @@ public enum BannerMultiMessageType: Int {
 public struct BannerMessageBaseStyle: BannerMessageStyle {
     public func makeBody(_ configuration: BannerMessageConfiguration) -> some View {
         VStack(spacing: 0) {
-            configuration.topDivider.frame(height: 4)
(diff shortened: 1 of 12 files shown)


Sources/FioriSwiftUICore/_FioriStyles/BannerMessageStyle.fiori.swift
```diff
@@ -90,17 +90,17 @@ extension BannerMessageFioriStyle {
         let bannerMessageConfiguration: BannerMessageConfiguration
         
         func makeBody(_ configuration: IconConfiguration) -> some View {
-            configuration.icon
-                .foregroundStyle(BannerMessageFioriStyle.titleForegroundColor(type: self.bannerMessageConfiguration.messageType))
+            Icon(configuration)
```
dyongxu, line 93: Why we change the default color to be negativeLabel , for icon, title, etc?
Originally, the default color is based on messageType, which looks good to me.
I think, to allow the customization, the enhancement is to allow developer to override the default color for different messageType.

dyongxu (changes requested):
some question inlined.

Sources/FioriSwiftUICore/_FioriStyles/BannerMessageStyle.fiori.swift
hengyi-zhang: It's the default logic before feedback patterns. I just restore the code to the node before feedback patterns.

billzhou0223 (approved):

dyongxu (approved):

PR #830, opened 2024-10-22

hengyi-zhang:


dyongxu (approved):

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
billzhou0223, line 1: Why does this public component not built with component protocols?

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
```diff
@@ -248,6 +248,14 @@ public struct BannerMultiMessageSheet: View {
                                 }, bannerTapAction: {
                                     self.showItemDetail(category: element.category, at: message.id)
                                 }, alignment: .leading, hideSeparator: true, messageType: message.messageType)
+                                    .ifApply(true, content: { v in
```
billzhou0223, line 251: Why use `ifApply` if the condition is always true?

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
```diff
@@ -248,6 +248,14 @@ public struct BannerMultiMessageSheet: View {
...
+                                        v.iconStyle { c in
+                                            c.icon.foregroundStyle(BannerMessageFioriStyle.titleForegroundColor(type: message.messageType))
```
billzhou0223, line 253: Customizing style like this will make style composition not work. Better way is 

`Icon(configuration).foregroundStyle(...)`

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
hengyi-zhang: Banner message sheet pops from the top banner message and nested under the title of the top banner message. 
Banner message sheet reuses several banner message items. In view of the little components in the top banner message use global universal environment key, like \.iconStyle, \.titleStyle, etc. The style of the top banner message will affect the banner message's style in the sheet. So here uses the .ifApply modifier to get the configuration of each banner message to reconfiguration the style which affected by top banner message.

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
billzhou0223: You can still build `BannerMessageSheet` with the new pattern even if a sub-component is a legacy component. You just need to clearly define the properties of `BannerMessageSheet` in protocols. You are not even exposing any properties of dimension selector in this component so it shouldn't affect you at all.

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
billzhou0223: I am not sure if the banner sheet is a subview of top banner. It looks weird to me. Anyway, I am wondering why you use `ifApply` to configure the icon style instead of doing something simpler. Aren't they the same?

```
BannerMessage()
    .iconStyle(...)
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
hengyi-zhang: fixed!

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
hengyi-zhang: fixed!

Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheet.fiori.swift
hengyi-zhang: fixed!

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -631,3 +631,24 @@ protocol _ToastMessageComponent: _IconComponent, _TitleComponent {
...
+    /// callback when this component want to dismiss itself
+    var dismissAction: (() -> Void)? { get }
+    /// callback when category or single item is removed
+    var removeAction: ((String, UUID?) -> Void)? { get }
+    /// callback when the link button is clicked
+    var viewDetailAction: ((UUID) -> Void)? { get }
+    // sourcery: defaultValue = true
+    /// the mark to turn on section header or not
+    var turnOnSectionHeader: Bool { get }
+    // sourcery: @ViewBuilder
```
billzhou0223, line 646: The type is already a closure, so the annotation should be `@ViewBuilder` instead of `// sourcery: @ViewBuilder`.

The template needs to be adjusted accordingly.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
hengyi-zhang: fixed!

billzhou0223 (approved):
Thanks

dyongxu (approved):

PR #834, opened 2024-10-24
feat: 🎸 [HCPSDKFIORIUIKIT-2791]SwiftUI: floating value support

zzchao-1999:
Screenshot 2024-10-18 at 17 14 37


Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -133,10 +133,14 @@ protocol _JouleWelcomeScreen: _MediaImageComponent, _GreetingTextComponent, _Tit
 protocol _StepperFieldComponent: _DecrementActionComponent, _TextInputFieldComponent, _IncrementActionComponent {
     /// The step value
     // sourcery: defaultValue = 1
-    var step: Int { get }
+    var step: Double { get }
     
     /// a range of values
-    var stepRange: ClosedRange<Int> { get }
+    var stepRange: ClosedRange<Double> { get }
```
billzhou0223, line 139: How do I control the interval of the stepper?

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
zzchao-1999: Did you mean the increment/decrement size? If so, you can use the "step" parameter.
var step: Double { get }

zzchao-1999:
> We may need to enhance the display of number of decimals in the StepperField. e.g. if you have a range of 0 to 1.0, and stepper value is 0.1, when you increase/decrease, you may see values as "0.6999999999999". This is due to "Double" is used as the new data type. One suggestion could be, we can check the stepper value. Then use the same number of decimals, to format the StepperField value

Added a number formatter to ensure the display of number of decimals in the stepperField.

dyongxu (approved):

billzhou0223 (approved):

dyongxu:
@zzchao-1999 Check with the reporter if it is needed for rel-4.1

PR #837, opened 2024-10-25
fix: 🐛 [JIRA:IOSSDKBUG-418] Fix FilterFeedbackBar layout

hengyi-zhang:


dyongxu (approved):
@hengyi-zhang Need a cherrypick to rel-4.1

PR #843, opened 2024-10-28
refactor: 💡 [JIRA: IOSSDKBUG-402] use styleable tag templete

xiaoyu0722:
|light|dark|
|-|-|
|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-28 at 17 17 26|Simulator Screenshot - iPhone 16 Pro Max - 2024-10-28 at 17 17 33|

### DIFF
--- a/Sources/FioriSwiftUICore/Utils/ConfigurationViewWrapper.swift
+++ b/Sources/FioriSwiftUICore/Utils/ConfigurationViewWrapper.swift
@@ -4,10 +4,6 @@ import SwiftUI
 public struct ConfigurationViewWrapper: View {
     var v: any View
     
-    init(_ view: some View) {
-        self.v = view
-    }
-    
     init(_ view: any View) {
         self.v = view
     }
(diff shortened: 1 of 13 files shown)


Sources/FioriSwiftUICore/Utils/ConfigurationViewWrapper.swift
```diff
@@ -8,6 +8,10 @@ public struct ConfigurationViewWrapper: View {
         self.v = view
     }
     
+    init(_ view: any View) {
+        self.v = view
+    }
```
billzhou0223, line 13: Why do you add this init?

xiaoyu0722: I'm also confused that compiler failed without this.
Screenshot 2024-10-29 at 07 22 16

xiaoyu0722: I'm not sure if it's caused by system's api.
`nonisolated public func tag<V>(_ tag: V, includeOptional: Bool = true) -> some View where V : Hashable`

billzhou0223: Not sure. I think we should keep only one of the init. If `any View` work fine, you can remove the other init.

billzhou0223 (approved):

PR #846, opened 2024-10-28
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2780] Progress Indicator

angiexyang:


KevinZK:
1.`_LinearProgressIndicatorComponent` this basic component, you only used the variable `indicatorProgress` inside.A variable can be added to `_ProgressIndicatorViewComponent`,Because `ProgressView() `is not used in ProgressIndicatorViewStyle.

Or 

The way you implement Circular progress in the ProgressIndicatorViewStyle file is through `Circle()`. Perhaps you create a new component for Circular progress. 
```
protocol _CircularProgressComponent {
    // sourcery: @Binding
    var progress: Double { get }
}

protocol _ProgressIndicatorViewComponent: _CircularProgressComponent, _TitleComponent{...}
```
In this way, you can assemble them in the ProgressIndicatorViewStyle file. When developers use it, they can also customize the style of Circular progress.

2.It seems that a `_TitleComponent` protocol is still missing because in the SAPFiori for UIKit project, textLabel is supported. 
“Text label below the circular activity indicator.”

3. In the SAPFiori UIKit project, `FUIModalProcessingIndicatorView` supports setting BackgroundMode.
```
/// The visual style of the background overlay behind the processing indicator.
    public enum BackgroundMode {
        /// The translucent white background.
        case translucentWhite
        /// The extra-light blurred background.
        case blurred
    }
/// The appearance of the background overlay behind the processing indicator. The default is `.translucentWhite`.
    public var backgroundMode: BackgroundMode = .translucentWhite

```

angiexyang:
> 1.`_LinearProgressIndicatorComponent` this basic component, you only used the variable `indicatorProgress` inside.A variable can be added to `_ProgressIndicatorViewComponent`,Because `ProgressView() `is not used in ProgressIndicatorViewStyle.
> 
> Or
> 
> The way you implement Circular progress in the ProgressIndicatorViewStyle file is through `Circle()`. Perhaps you create a new component for Circular progress.
> 
> ```
> protocol _CircularProgressComponent {
>     // sourcery: @Binding
>     var progress: Double { get }
> }
> 
> protocol _ProgressIndicatorViewComponent: _CircularProgressComponent, _TitleComponent{...}
> ```
> 
> In this way, you can assemble them in the ProgressIndicatorViewStyle file. When developers use it, they can also customize the style of Circular progress.
> 
> 2.It seems that a `_TitleComponent` protocol is still missing because in the SAPFiori for UIKit project, textLabel is supported. “Text label below the circular activity indicator.”
> 
> 3. In the SAPFiori UIKit project, `FUIModalProcessingIndicatorView` supports setting BackgroundMode.
> 
> ```
> /// The visual style of the background overlay behind the processing indicator.
>     public enum BackgroundMode {
>         /// The translucent white background.
>         case translucentWhite
>         /// The extra-light blurred background.
>         case blurred
>     }
> /// The appearance of the background overlay behind the processing indicator. The default is `.translucentWhite`.
>     public var backgroundMode: BackgroundMode = .translucentWhite
> ```

BackgroundMode and textLabel are supported in FUIProcessingIndicator and FUIModalProcessingIndicator, but not FUIProgressIndicator. I'll look into creating a CircularProgressIndicator base component which can be used in both progress indicator and processing indicator.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -576,3 +576,19 @@ protocol _ToastMessageComponent: _IconComponent, _TitleComponent {
...
+///
+/// ## Usage
+/// ```swift
+/// @State var progress: Double = 0.0
+/// @State var progressState = ProgressViewState.processing
+///
+/// ProgressIndicatorView(indicatorProgress: self.$progress, progressState: self.$progressState)
+/// ```
+// sourcery: CompositeComponent
+protocol _ProgressIndicatorViewComponent: _LinearProgressIndicatorComponent {
```
billzhou0223, line 590: I'll name it `ProgressIndicator` instead of `ProgressIndicatorView`.

Do not conform to `_LinearProgressIndicatorComponent` whose name is not appropriate here. You can instead create a base component `_ProgressIndicatorProtocol` with a single property `progress: Progress`.

Sources/FioriSwiftUICore/_FioriStyles/ProgressIndicatorViewStyle.fiori.swift
```diff
@@ -0,0 +1,138 @@
+import FioriThemeManager
+
+// Generated using Sourcery 2.1.7 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+import Foundation
+import SwiftUI
+
+public enum ProgressViewState {
```
billzhou0223, line 8: If I understand correctly, each state represents a different UI style. I will remove this state definition and create custom `ProgressIndicatorStyle` implementations for each style. From users' perspective, the code should look like this

```
ProgressIndicator(progress: 0.3)
   .progressIndicatorStyle(ProgressIndicatorPausableStyle(isPaulsed: false, icon: { isPaused in ... }))

ProgressIndicator(progress: 0.3)
   .progressIndicatorStyle(ProgressIndicatorStoppableStyle(isPaulsed: false, icon: { isPaused in ... }))
```

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
angiexyang: There's an existing `ProgressIndicator` used in `InfoView`, do you have another naming suggestion?

billzhou0223: If it's already existed, introducing a new component called `ProgressIndicatorView` doesn't make sense to me. We need to consolidate them. I recommend removing the old one.

dyongxu: @angiexyang After our mtg, I looked into this as well. ProgressIndicator/ProgressIndicatorView were implemented with "old" sourcery , which is defined in ModelDefinition.swift. So, this part of the code needs a re-write anyway to use the "new" sourcery to support the component styles. It makes sense to keep naming in your PR.

FYI:  some component re-write already happened earlier. e.g. ListPicker, SideBar, etc. You can refer to ModelDefinition.swift to see how it was done for other component.

e.g. @available(*, unavailable, renamed: "_ListPickerItemModel", message: "Will be removed in the future release. Please create ListPickerItem with other initializers instead.")
public protocol ListPickerItemModel {}

PR #854, opened 2024-10-31
fix: 🐛 [JIRA:IOSSDKBUG-375] TextInput Fom View a11y fix 2

janhuachu:
✅ Closes: IOSSDKBUG-375


Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -123,6 +123,8 @@ protocol _TextFieldFormViewComponent: _TitleComponent, _TitleFormViewComponent,
     var actionIcon: Image? { get }
     /// The action to be performed when the action button is tapped.
     var action: (() -> Void)? { get }
+    /// The accessibility label for the action icon
+    var actionIconAccessibilityLabel: String? { get }
```
billzhou0223, line 127: My thinking:

1. TextFieldFormView should conform to `ActionComponent` instead of exposing a `actionIcon` property.
2. The accessibility label should be applied to action button instead of action icon. And developers can set it using `actionStyle` modifier so we don't need to expose this property. Providing accessibility label property in a component is not scalable and maintainable.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
janhuachu: I will try to make it ActionComponent. However, when I added Button to the view, my experience is that the whole view become the touch area for the Button that user could not edit the TextField. I think this is SwiftUI problem.

billzhou0223: I believe the button will only appear on the trailing side of the field so it should not cover the entire input field.

janhuachu: I originally use SwiftUI's Button and I observe the behavior that the entire view is the touch area of the Button. I need to chage it to an Image and add a onTapGesture to get the correct behavior.

When I use ActionComponent, I need to get the action block from the ActionComponent to invoke the action. I have yet to be able to get the action block from the ActionComponent. Do you have any idea how to do that?

janhuachu: Just tried a simple example:

`    var body: some View {
        List {
            HStack {
                Text("Test")
                Button("Button") {
                    print("*** Hello World!")
                }
            }
        }
    }
`
When tap "Text" the button action is also triggered.

billzhou0223 (approved):

PR #867, opened 2024-11-06

hengyi-zhang:
1. Replace example icons
2. Dismiss banner message when modal sheet appear


billzhou0223 (approved):

dyongxu (changes requested):
1. I have talked with Grace. The dismiss of "Popover" on iPad does not look right. (The arrow is anchored to the "Bar Title". It does not make sense). Grace will discuss with Chenchen .
2. I noticed you changed the icon in the Example App.  Do we have a default value for the icon in SDK code? I thought SDK should have default value if App developer does not provide one.
3. I happened to notice, in the row with the "Clear" button, you can click anywhere in this row and Clear action is triggered. I think it is a wrong behavior, usually it should be triggered only if you tap the Clear button

dyongxu (approved):

PR #878, opened 2024-11-12
fix: 🐛 [IOSSDKBUG-455] FilterFeedbackBar maxWidth customization

restaurantt:
Add 'maxWidth' in 'filterFeedbackBarStyle'. If width is not enough, text will be truncated.


dyongxu (changes requested):
Add a test case in TestApp, to show this behavior

billzhou0223 (approved):

restaurantt:
turn on the Custom Style switch at the bottom, it will active the custom style.
Simulator Screenshot - iPhone 16 Pro Max - 2024-11-13 at 08 58 26

dyongxu (approved):

PR #885, opened 2024-11-14
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2828] PreviewDevice is ignored in a #Preview macro

JunSong-SH:
[HCPSDKFIORIUIKIT-2828](https://jira.tools.sap/browse/HCPSDKFIORIUIKIT-2828) PreviewDevice is ignored in a #Preview macro

This PR handle the change to clean up the warning message 'PreviewDevice is ignored in a #Preview macro' with XCode 16.1. 

See the following [Xcode 15 release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Previews):
_**The preview canvas has a new control for picking which device to use for previewing. By default it tracks the device family of the selected run destination, but a specific device can be selected. This should be used in favor of the previewDevice modifier.**_

Therefore, going forward, the `previewDevice` modifier should not be used. Instead, the control on the canvas GUI should be used to select a preview device.

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/KPI/KPIProgressViewExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/KPI/KPIProgressViewExample.swift
@@ -28,7 +28,6 @@ struct KPIProgressPreview: PreviewProvider {
     static var previews: some View {
         Group {
             KPIProgressExample()
-                .previewDevice(PreviewDevice(rawValue: "iPhone 12"))
                 .previewDisplayName("iPhone 12")
         }
     }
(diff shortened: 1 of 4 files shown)


dyongxu (changes requested):
I think there are quite a few other files with the same warning.
Can we combine the fixes for other files in the same PR?

JunSong-SH:
> I think there are quite a few other files with the same warning. Can we combine the fixes for other files in the same PR?

Double check it and search for more similar code using the keyword. But, there is no warning because they are not in #Preview macro. Now, all code should be clear. Thanks!

dyongxu (approved):

PR #892, opened 2024-11-19
fix: 🐛 [IOSSDKBUG-447]SortFilterItem Reset button func

restaurantt:
SortFilterItem.PickerItem Reset button functionality. 
1. Provide reset button configuration to customize hidden, title, type.
2. Selected section add deselect all.
3. Selected in header add count 'Selected (2)' to show, default is showing count.
4. update select all showing logic.
Simulator Screenshot - iPhone 16 Pro Max - 2024-11-20 at 11 07 26
Simulator Screenshot - iPhone 16 Pro Max - 2024-11-20 at 11 07 06

### DIFF
--- a/Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
+++ b/Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
@@ -78,15 +78,20 @@ struct CancellableResettableDialogNavigationForm<Title: View, CancelAction: View
     }
     
     var body: some View {
+        let isNotIphone = UIDevice.current.userInterfaceIdiom != .phone
         NavigationStack {
-            VStack(spacing: UIDevice.current.userInterfaceIdiom != .phone ? 8 : 16) {
-                #if !os(visionOS)
-                    self.components.background(Color.preferredColor(.secondaryGroupedBackground))
-                #else
-                    self.components.background(Color.clear)
-                #endif
-                self.applyAction
-                    .accessibilityIdentifier("Apply")
+            ZStack {
+                Color.preferredColor(.chromeSecondary)
+                    .ignoresSafeArea()
+                VStack(spacing: isNotIphone ? 8 : 16) {
+                    self.components
+                    
+                    VStack(spacing: 0) {
+                        self.applyAction
+                            .accessibilityIdentifier("Apply")
+                        Spacer().frame(height: isNotIphone ? 13 : 16)
+                    }
+                }
             }
             .navigationBarTitleDisplayMode(.inline)
             .toolbar {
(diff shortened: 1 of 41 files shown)


Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -346,6 +346,37 @@ extension SortFilterItem {
...
+    var title: String
+    var isHidden: Bool
+    
+    init(type: FilterFeedbackBarResetButtonType = .default, title: String, isHidden: Bool = false) {
+        self.type = type
+        self.title = title
+        self.isHidden = isHidden
+    }
+    
+    /// Default FilterFeedbackBarResetButtonConfiguration
```
xiaoyu0722, line 361: Convenience initialization for `reset` button type.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -346,6 +346,37 @@ extension SortFilterItem {
     }
 }
 
+/// FilterFeedbackBar ResetButton Configuration
+public struct FilterFeedbackBarResetButtonConfiguration: Equatable {
```
xiaoyu0722, line 350: Name confusion, you can try `FilterFeedbackRightBarButtonConfiguration`.

dyongxu (changes requested):
In the test case for "List Multiple", click "Select All", then click it again, all selected values are cleared.  I would expect the second click will not trigger any clear action.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
dyongxu: @xiaoyu0722 I think we should not use the wording "right" or "left" in the API names, since it will be confusing in the "right to left" language. It looks ok to me for the current name, but pls double check with Jiajun if that is actually what he wants

xiaoyu0722: If it's a `reset` button, but it will also work for clear all, that's what makes me confused.

Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
```diff
@@ -27,13 +27,14 @@ struct CancellableResettableDialogForm<Title: View, CancelAction: View, ResetAct
...
                 self.title
-                Spacer()
-                self.resetAction.accessibilityIdentifier("Reset")
-            }
+                HStack {
+                    self.cancelAction.accessibilityIdentifier("Cancel")
+                    Spacer()
+                    self.resetAction.accessibilityIdentifier("Reset")
+                }
+            })
```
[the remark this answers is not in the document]
restaurantt: Haven't find a good solution for now. Try later.

restaurantt:
> In the test case for "List Multiple", click "Select All", then click it again, all selected values are cleared. I would expect the second click will not trigger any clear action.

Fixed.

xiaoyu0722 (approved):

dyongxu (approved):

PR #927, opened 2024-12-05
fix: 🐛 [IOSSDKBUG-485]Filter layout issue for very large fonts

restaurantt:
Use NavigationStack instead of HStack, adapt to dynamic type changing.


Sources/FioriSwiftUICore/Views/OptionListPickerItem+View.swift
```diff
@@ -22,6 +22,24 @@ public enum OptionListPickerItemLayoutType {
     case flexible
 }
 
+public extension OptionListPickerItem {
+    /// create a filter picker which used in FilterFeedbackBarItem
+    /// - Parameters:
+    ///   - value: Selected value indexs.
```
dyongxu, line 28:
```suggestion
    ///   - value: Selected value index.
```

dyongxu (commented):
doc update

Sources/FioriSwiftUICore/Views/OptionListPickerItem+View.swift
restaurantt: value is [Int] type. Should it be plural?

restaurantt: Change to 'Indexes for selected values.'

dyongxu (approved):

PR #928, opened 2024-12-06
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2880] ActivityItemModel Refactor

hengyi-zhang:


dyongxu (changes requested):
I saw in the FioriIcons/Resources, some files were updated, (Contents.json).
Usually, we don't update these files under Resource folder, since these files are delivered by Designer team.  
Is there some specific reason why we have to update them here ?

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -726,3 +728,10 @@ protocol _ProgressIndicatorComponent: _ProgressIndicatorProtocol {}
 /// ```
 // sourcery: CompositeComponent
 protocol _ProcessingIndicatorComponent: _OptionalTitleComponent {}
+
```
angiexyang, line 731: Could you add some documentation and example code usage?

hengyi-zhang:
> I saw in the FioriIcons/Resources, some files were updated, (Contents.json). Usually, we don't update these files under Resource folder, since these files are delivered by Designer team. Is there some specific reason why we have to update them here ?

fixed!

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
hengyi-zhang: fixed!

xiaoqinggrace:
For the generated code, we do not need to commit if there is no change even if the order of the generated code is different.

Sources/FioriSwiftUICore/Models/ModelDefinitions.swift
```diff
@@ -97,7 +97,11 @@ public protocol ActionModel: ActionComponent {}
 public protocol TextInputModel: TextInputComponent {}
 
 // sourcery: generated_component
-public protocol ActivityItemModel: IconComponent, SubtitleComponent {}
+public protocol _ActivityItemModel: IconComponent, SubtitleComponent {}
+
+/// Deprecated ActivityItemModel
+@available(*, unavailable, renamed: "_ActivityItemModel", message: "Will be removed in the future release. Please create ActivityItemModel with other initializers instead.")
```
dyongxu, line 103:
```suggestion
@available(*, unavailable, renamed: "_ActivityItemModel", message: "Will be removed in the future release. Please create ActivityItem with other initializers instead.")
```

billzhou0223 (approved):

KevinZK (approved):

Sources/FioriSwiftUICore/Models/ModelDefinitions.swift
hengyi-zhang: fixed!

dyongxu (approved):

PR #931, opened 2024-12-09
fix: 🐛 [IOSSDKBUG-482]iPad popover FilterFeedback list jumping

restaurantt:
Select/Deselect All cause view hight change.
On iPad, popover with UIKit set 'preferredContentSize'.
On iPhone, add animation to list.


billzhou0223 (changes requested):
Please add a video demonstrating the result.

dyongxu (changes requested):
The click event will be only triggered in the center of the bar item. It does not work in the area close to the border

restaurantt:
> The click event will be only triggered in the center of the bar item. It does not work in the area close to the border

Fixed.

restaurantt:
> Please add a video demonstrating the result.


https://github.com/user-attachments/assets/73899944-2c36-4628-86e4-1f5df286c9cf

https://github.com/user-attachments/assets/db1958fd-0b3d-4a05-804a-27514318442a

1. iPad: popover has animation now. Tried to add animation to list and apply button, but performance is not so good. Also tried add animation to height changing, not work. 
2. iPhone, when 'presentationDetents' use '.height', looks not so good. Other cases are ok.

dyongxu (approved):

dyongxu:
@restaurantt Since this looks better than before, I have approved it for this PR.  We can continue to look into the improvement.

billzhou0223 (approved):

PR #932, opened 2024-12-09
fix: 🐛 avoid conflict btw drag and tap gesture in Carousel

shengxu7:
Avoid going to the detail view during dragging a carousel in Joule


dyongxu (changes requested):
Build failed due to extra argument isEnabled: true

dyongxu (approved):

PR #935, opened 2024-12-11
refactor: 💡 Added NSS  support

KevinZK:
## Example
```swift

/// file:FioriDesignLanguage.nss
FioriLoadingIndicator_progress {
    tint-color: red;
}

/// Support setting styles for different IDs.
fiori_loadingindicator_component_0_progress {
    tint-color: red;
    control-size: regular;
}

fiori_loadingindicator_component_0_title {
    font-weight: bold;
    font-size: 18;
    font-color: green;
}

fiori_loadingindicator_component_0_content {
    border-color: yellow;
    position-layout: bottom;
}

// file:FioriLoadingIndicatorView.nss
@testColor0:#A68A5BFF;
@testColor1:green;

fiori_loadingindicator_component_progress {
    tint-color: yellow;
    control-size: large;
}

fiori_loadingindicator_component_title {
    font-weight: bold;
    font-size-selected: 48;
    font-color: @testColor0;
    font-color-selected: black;
    font-color-disabled: @testColor0;
}

fiori_loadingindicator_component_content {
    border-color: red;
    position-layout: trailing;
}

/// appDegelate
NSSTool.registerNSSStyleSheet("FioriDesignLanguage")

// LoadingIndicatorExample.swift
LoadingIndicator(title: "Loading...", isPresented: self.$isPresented1).loadingIndicatorStyle(.nss("FioriLoadingIndicatorView"))

/*
If a component's style is set in the global style sheet and registered, then the global style applies the component
*/

```
Taking LoadingIndicator as an example, we can split LoadingIndicator into three sub-components: content, title, and progress. After parsing the nss file, set NSS styles for these three sub-components respectively.

In the code template, two styles, "base" and "fiori", have already been defined. We can extend more styles as needed. So I have added support for the "NSS" style.
```swift
public extension LoadingIndicatorStyle where Self == LoadingIndicatorNSSStyle {
    static func nss(_ parserType: NSSParserType) -> LoadingIndicatorNSSStyle {
        parserType.mergeNSSData()
        return LoadingIndicatorNSSStyle()
    }

    static var globalNSS: LoadingIndicatorNSSStyle {
        LoadingIndicatorNSSStyle(isGlobal: true)
    }
}
```
`configuration.titleIdentifier`, `configuration.progressIdentifier`, and `configuration.contentIdentifier` define unique identifiers for the sub-components of the LoadingIndicator component in the `LoadingIndicatorConfiguration` file and are used to identify styles in nss.

### Global NSS style
```swift
extension EnvironmentValues {
    var loadingIndicatorStyle: any LoadingIndicatorStyle {
        self.loadingIndicatorStyleStack.last ?? .base.concat(.fiori).concat(.globalNSS)
    }
}

```
Manage the style stack. The highest priority is the nss style. If a global style is set, then the component will preferentially use the nss style.

`NSSTool.registerNSSStyleSheet("FioriDesignLanguage")`

###  Custom style types
```swift
enum LoadingIndicatorNSSType: String, NSSCovert {
    case positionLayout = "position-layout"
    
    func apply(value: Any, to view: any View) -> any View {
        switch self {
        case .positionLayout:
            if let name = value as? String {
                return view.indicatorPosition(getPosition(name))
            }
        }
        return view
    }

    func getPosition(_ name: String) -> LoadingIndicator.Layout {
        switch name {
        case "leading":
            return .leading
        case "trailing":
            return .trailing
        case "top":
            return .top
        case "bottom":
            return .bottom
        default: return .top
        }
    }
}

// Usage:
struct ContentNSSStyle: LoadingIndicatorStyle {
        let loadingIndicatorConfiguration: LoadingIndicatorConfiguration
        let nssData: NSSStyleData
        func makeBody(_ configuration: LoadingIndicatorConfiguration) -> some View {
            LoadingIndicator(configuration)
                .modifier(NSSStyleModifier<NSSBaseStyleType>(styles: nssData))
                .modifier(NSSStyleModifier<LoadingIndicatorNSSType>(styles: nssData))
        }
    }
```
The `LoadingIndicator` component has a `position` variable used to change the position of the icon and text. Four positions are defined through the enumeration `LoadingIndicatorNSSType`, and the `NSSCovert` protocol is implemented. The `apply` method converts the style value into the `LoadingIndicator.Layout` type.

###  Support different states

Currently supports six states: `normal`,` selected`, `highlighted`, `disabled`, `readonly`, and `focused`. Other states can be added.

The parsing tool will default to parsing styles with these six suffixes and select the final style according to the state.

### Discussion

1.  Currently, it's not possible to synchronize the classPaths of the SwiftUI SDK and the UIKit SDK.
There are significant differences in the ways they build components, which results in the inability to make the classPaths of child components consistent.

2.  API differences
Parameters of the background API: Background. This is a generic constraint, indicating that the Background type must conform to the View protocol.
When writing an NSS file, it is usually written like this:` background - color: blue`. In the default conversion function, only the color case is handled. Of course, we can also customize special conversions.
However, during the process of designing the style of components, some special processing will be done on the background:
`.background(RoundedRectangle(cornerRadius: 8).fill(Color.green))`
Here, the background is set with a corner radius of 8 and a fill color of green.
In this case, `background - color: blue` cannot meet this style. Two methods can solve this problem:
   - Split the existing API into:
`.background(.green).cornerRadius(8)`
   - Use custom style types to perform special processing on `background - color: blue`.
  
3.  How to allow all child components to independently set NSS styles in a multi - nested component scenario?

The current template architecture cannot perfectly solve this problem. I've already provided a solution in #959  .
I've assigned an ID to each component and extended the Configuration with an `isDirectChild(_ componentIdentifier: String) -> Bool` method to determine the affiliation of child components.
  

## Overview
NSS is a style system similar to CSS, which allows developers to define and manage the styles of SwiftUI views through style sheets. This system supports dynamic style application, style inheritance, and reuse.

### 1. NSSManager
The core management class responsible for loading and managing style sheets.

Main Functions:
- Style cache management
- Global style management
- Style sheet file loading
- Style merging processing

### 2. NSSParser
The parser responsible for parsing .nss files.

Main Functions:
- Comment processing
- Variable parsing
- Style block parsing
- Attribute value conversion

### 3. NSSStyleModifier
A modifier used to apply styles to SwiftUI views.

### 4. NSSCovert Protocol
Defines the standard interface for style conversion.

### Effect screenshot

Simulator Screenshot - iPhone 16 Pro - 2025-01-16 at 17 17 35


Apps/Examples/Examples/FioriSwiftUICore/Indicator/LoadingIndicatorExample.swift
```diff
@@ -31,6 +31,18 @@ struct LoadingIndicatorExample: View {
                         .fontWeight(.bold)
                 }.padding(.bottom, 10)
                 
+                Section {
+                    LoadingIndicator(title: "Loading...", isPresented: .constant(true))
+                        .loadingIndicatorStyle(.nss("FioriLoadingIndicatorView"))
```
xiaoyu0722, line 36: Is there competition between styles. Like below example, will the nss still be respected?
```swift
LoadingIndicator(title: "Loading...", isPresented: .constant(true))
  .titleStyle { c in
    c.title.foregroundStyle(Color.red)
  }
  .loadingIndicatorStyle(.nss("FioriLoadingIndicatorView"))
```

CharlesXu0488:
I barely remember nss in UIKit SAPFiori can customize images. Are we going to support image?
What is the relationship between default Fiori style and nss + nss style? one or the other?

KevinZK:
> I barely remember nss in UIKit SAPFiori can customize images. Are we going to support image? What is the relationship between default Fiori style and nss + nss style? one or the other?

Whether to support image is decided by oneself when developing components.
I think
The "nss style" is a theme loaded from the nss file.Kind of like css
nss is a file with the extension.nss.
The Fiori style should be a style set according to the designer's design scheme.

CharlesXu0488:
> > I barely remember nss in UIKit SAPFiori can customize images. Are we going to support image? What is the relationship between default Fiori style and nss + nss style? one or the other?
> 
> Whether to support image is decided by oneself when developing components. This loading icon is provided by the system component progressView(). The changes in size and color are controlled by tintcolor and control size. I think The "nss style" is a theme loaded from the nss file.Kind of like css nss is a file with the extension.nss. The Fiori style should be a style set according to the designer's design scheme.

What if putting nss behind nss? If Fiori style sees nss, uses nss; otherwise falls back to default Fiori style?
Can you provide an image example?

KevinZK:
> > > I barely remember nss in UIKit SAPFiori can customize images. Are we going to support image? What is the relationship between default Fiori style and nss + nss style? one or the other?
> > 
> > 
> > Whether to support image is decided by oneself when developing components. This loading icon is provided by the system component progressView(). The changes in size and color are controlled by tintcolor and control size. I think The "nss style" is a theme loaded from the nss file.Kind of like css nss is a file with the extension.nss. The Fiori style should be a style set according to the designer's design scheme.
> 
> What if putting nss behind nss? If Fiori style sees nss, uses nss; otherwise falls back to default Fiori style? Can you provide an image example?

In the current template, In the case of reusing the nss style, only the first style takes effect.
I am not sure whether this design was originally to solve a certain problem.So I did not modify this logic.

If the parsed data in the nss file is empty, then the Fiori style still takes effect.

KevinZK:
There are too many conflicts. Submit a new PR after fixing them.

PR #946, opened 2024-12-25
refactor: 💡 [JIRA:HCPSDKFIORIUIKIT-2880] ActivityItems Refactor

hengyi-zhang:


JunSong-SH (approved):
LGTM!

xiaoyu0722 (commented):
I don't think we need add all cases for these sub-component.
We can group them into it's use case if you really want it, like move them into `Contact Item` case.
Also, what's the difference between `ActivityItems` and `ActionItems`.
Seems they are the same thing. Why don't we just use `ActivityItems` directly? You can double check with David.

hengyi-zhang:
ActivityItems is the same as actionItems, ActivityItems will replace the former actionItems.

PR #950, opened 2024-12-30
feat: 🎸 [HCPSDKFIORIUIKIT-2766] FilterFeedbackBar Range Slider

restaurantt:
Support Range Slider with basic function.
Simulator Screenshot - iPhone 16 Pro Max - 2024-12-30 at 10 08 00
Simulator Screenshot - iPad Pro 13-inch (M4) - 2024-12-30 at 10 06 50


Sources/FioriSwiftUICore/Views/SortFilter/_SortFilterCFGItemContainer.swift
```diff
@@ -298,12 +298,22 @@ extension _SortFilterCFGItemContainer: View {
...
             }
-            SliderPickerItem(
-                value: Binding<Int?>(get: { self._items[r][c].slider.workingValue }, set: { self._items[r][c].slider.workingValue = $0 }),
-                formatter: self._items[r][c].slider.formatter,
-                minimumValue: self._items[r][c].slider.minimumValue,
-                maximumValue: self._items[r][c].slider.maximumValue
-            )
+            var sliderItem = self._items[r][c].slider
+            if sliderItem.sliderMode == .single {
+                FioriSlider(
```
JunSong-SH, line 303: The minimumValue and maximumValue should be set as a range both for Single and Range slider. I would suggest that set the 'step' to Slider. It means that above 'SortFilterItem' need expose new optional parameter 'step' with default 1.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -640,41 +656,107 @@ public extension SortFilterItem {
...
             }
-            return self.name
         }
         
         mutating func setValue(newValue: SliderItem) {
-            self.value = newValue.value
+            if self.sliderMode == .single {
+                self.value = newValue.value
+            } else {
+                self.lowerValue = newValue.lowerValue
```
JunSong-SH, line 741: As the default range slider allow user to edit the value in text field directly, so, I wound think we can add a validation here simply to let the editing lower value and upper value always in range(minimumValue, maximumValue) to avoid the unexpected value was return to filter item.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -623,12 +623,28 @@ public extension SortFilterItem {
         public var value: Int?
         var workingValue: Int?
         let originalValue: Int?
+        
+        public var lowerValue: Int?
```
JunSong-SH, line 627: I am not sure why original Slider only allow Int value. But, now, as FioriSlider already support the decimal numbers, so, I would think we need Double value instead of Int value here in case any consumer need the decimal numbers.
In that case, we need value, workingValue, originalValue, minimumValue and maximumValue were Double type. Similarly, lowerValue and upperValue should be Double type. FioriSlider has a Int type parameter 'decimalPlaces' to indicate the supported decimal. By default, it was 0 means the value is playing as Int.
So, I would think we should have three 'init' method. One for backward compatibility and still use Int-type value, minimumValue and maximumValue. For other two, the Double-type value should be used for Single and Range Slider with 'step' double-type parameter(default value 1), 'decimalPlace'(default value 0).

Sources/FioriSwiftUICore/Views/SortFilter/_SortFilterCFGItemContainer.swift
restaurantt: Original api provide minimumValue and maximumValue individually, so I followed that.
I will add 'step'.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
restaurantt: Thank you. Let me improve that.

JunSong-SH: Talked with JiaJun and it seems like that the Figma design just need the Integer value for Slider Filter here. So, we can ignore the Double-type value supporting now.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -378,6 +378,31 @@ public struct FilterFeedbackBarResetButtonConfiguration: Equatable {
     }
 }
 
+/// FilterFeedbackBar slider item value change handler
+public struct SliderItemValueChange: Equatable {
```
JunSong-SH, line 382: I would suggest change the struct name to SliderValueChangeHandler.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -378,6 +378,31 @@ public struct FilterFeedbackBarResetButtonConfiguration: Equatable {
...
+    /// Slider description style type
+    public enum SliderItemValueChangeType {
+        case fiori
+        case error
+        case warning
+        case informational
+        case success
+    }
+    
+    public var handler: (Double, Double) -> (SliderItemValueChangeType, String)
```
JunSong-SH, line 392: I would suggest change the method name to 'onValueChange'

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -378,6 +378,31 @@ public struct FilterFeedbackBarResetButtonConfiguration: Equatable {
     }
 }
 
+/// FilterFeedbackBar slider item value change handler
+public struct SliderItemValueChange: Equatable {
+    /// Slider description style type
+    public enum SliderItemValueChangeType {
```
JunSong-SH, line 384: I would suggest to change the enum name to 'SliderInformationType'

JunSong-SH (approved):
LGTM!

zzchao-1999 (approved):

PR #957, opened 2025-01-07
[JIRA:HCPSDKFIORIUIKIT-2882] TextInput Deprecated

hengyi-zhang:
TextInput Deprecated

### DIFF
--- a/Sources/FioriSwiftUICore/_ComponentProtocols/BaseComponentProtocols.swift
+++ b/Sources/FioriSwiftUICore/_ComponentProtocols/BaseComponentProtocols.swift
@@ -485,6 +485,15 @@ protocol _ActivityItemsComponent {
 }
 
 // sourcery: BaseComponent
+protocol _TextInputComponent {
+    // sourcery: @Binding
+    // sourcery: defaultValue = ".constant("")"
+    var textInputValue: String { get }
+    // sourcery: no_view
+    var onCommit: (() -> Void)? { get }
+}
+
+// sourcery: BaseComponent
 protocol _LowerThumbComponent {
     // sourcery: @ViewBuilder
     // sourcery: defaultValue = "Circle()"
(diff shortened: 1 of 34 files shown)


Sources/FioriSwiftUICore/_ComponentProtocols/BaseComponentProtocols.swift
```diff
@@ -244,7 +244,12 @@ protocol _MoreActionOverflowComponent {
 // sourcery: BaseComponent
 protocol _TextInputFieldComponent {
     // sourcery: @Binding
+    // sourcery: defaultValue = ".constant("")"
```
xiaoyu0722, line 247: Why need a constant default value? I think this is required from developer.

Sources/FioriSwiftUICore/_generated/Components/Component+Protocols.generated.swift
```diff
@@ -253,3 +246,10 @@ public protocol ActionComponent {
 public protocol ProgressIndicatorComponent {
     var progressIndicatorText: String? { get }
 }
+
+public protocol TextInputComponent : AnyObject {
+	// sourcery: bindingPropertyOptional=.constant("")
+    var textInputValue: String { get set }
+	// sourcery: no_view
+    var onCommit: (() -> Void)? { get }
+}
```
xiaoyu0722, line 255: Why there is extra tab before the annotation?

xiaoyu0722 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/BaseComponentProtocols.swift
hengyi-zhang: fixed!

Sources/FioriSwiftUICore/_generated/Components/Component+Protocols.generated.swift
hengyi-zhang: It is generated file, and the comments display normal in Xcode, but display extra tab on the github, the whole file has the same question.

hengyi-zhang: There is a possible reason that a tab represents 8 white spaces here.

PR #958, opened 2025-01-10
fix: 🐛 fix customization issue on Range Slider

JunSong-SH:
This PR handle the change to fix an issue for customization when consumer trying to hide the leading and trailing accessory view. Here, we should always allow the lower and upper thumb display but the leading and trailing accessory view can be hide.


dyongxu (commented):
Do we have a UI test case for this , to make them invisible?

JunSong-SH:
> Do we have a UI test case for this , to make them invisible?

Thanks for the reminder! I add a demo case for this invisible case.

dyongxu (approved):

PR #965, opened 2025-01-17
feat: 🎸 [JIRA:HCPSDKFIORIUIKIT-2859] DimensionSelector (refactor)

angiexyang:
DimensionSelector refactoring in SwiftUI

Screenshot 2025-01-17 at 3 02 43 PM
Screenshot 2025-01-22 at 1 32 48 PM


Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -1120,3 +1120,48 @@ protocol _HeaderChartComponent: _TitleComponent, _SubtitleComponent, _TrendCompo
...
+///  @State var selectedIndex: Int? = 0
+///  @ObservedObject var stockModel = Tests.stockModels[0]
+///
+///  DimensionSelector(titles: segmentTitles, selectedIndex: $selectedIndex)
+///     .onChange(of: selectedIndex) {
+///            stockModel.indexOfStockSeries = selectedIndex ?? -1
+///    }
+///  ```
+// sourcery: CompositeComponent
+protocol _DimensionSelectorComponent: _TitlesComponent {
```
JunSong-SH, line 1138: I feel it is more like a Chip:https://experience.sap.com/fiori-design-android/chips/. I would suggest that to make the Segment stylable. Basically, the idea was to use two components: One for item/segment and another for list/collection.

For example, we use _DimensionSegmentComponent and _DimensionSelectorComponent here:
protocol _DimensionSegmentComponent {
// sourcery: @Binding
    /// Whether the item is selected or not
    var isSelected: Bool { get }
}
This component present each segment. You can build the segments with your code like 'Text(self.modelObject.titles[index]).padding(self.modelObject.titleInsets)' in this segment component. Maybe, it is easy for us to extend it future. For example, to display an icon for each segment. 

Furthermore, you might can add a callback,  var segment: (String) -> any View { get }, in DimensionSelectorComponent and let App team have chance to construct the segment and customize it. In that case, each of segment was stylable and the property 'segmentAttributes' can be removed.

The above declared base component , _TitlesComponent, seems can be removed here. Just let the DimensionSelectorComponent accept the input data. For example:  titles:[String]. For me, we don't need titleInsets, segmentAttributes anymore. For 'isEnable' property, I think it also should be removed if I understand you correctly. In SwfitUI, the @Environment(\.isEnabled) can be used.

All above mentioned name for component or property just for example. You might can consider to use a meaningful name.

JunSong-SH (changes requested):
Additional, you might can also refer the #951 to consider to use the ViewBuilder to build the segments.

Sources/FioriSwiftUICore/_FioriStyles/DimensionSegmentStyle.fiori.swift
```diff
@@ -0,0 +1,97 @@
...
+    @Published var segmentWidthMode: SegmentWidthMode
+     
+    var _maxSegmentWidth: CGFloat?
+    
+    // Initializes a `DimensionSegmentModelObject` with attributes from `DimensionSelector` and default title insets.
+    public init(segmentIndex: Int, _maxSegmentWidth: CGFloat?, configuration: DimensionSelectorConfiguration) {
+        self.titles = configuration.titles
+        self.segmentIndex = segmentIndex
+        self._maxSegmentWidth = _maxSegmentWidth
+        self.titleInsets = EdgeInsets(top: 6, leading: 6, bottom: 6, trailing: 6)
```
JunSong-SH, line 94: The 'titleInsets' was only used for Text on DemensionSegment with fixed value. So, I think you can remove this from the Observable Object and just set the fixed value to the Text.

Sources/FioriSwiftUICore/_FioriStyles/DimensionSegmentStyle.fiori.swift
```diff
@@ -0,0 +1,97 @@
...
+                    return SegmentAttributes(textColor: Color.preferredColor(.secondaryLabel), font: Font.fiori(forTextStyle: .subheadline), borderWidth: 0.33, borderColor: Color.preferredColor(.separator), backgroundColor: Color.preferredColor(.tertiaryFill))
+                }
+            }
+            return SegmentAttributes(textColor: Color.preferredColor(.secondaryLabel), font: Font.fiori(forTextStyle: .subheadline), borderWidth: 0.33, borderColor: Color.preferredColor(.secondaryFill), backgroundColor: Color.preferredColor(.tertiaryFill))
+        }
+    }
+}
+
+class DimensionSegmentModelObject: ObservableObject {
+    @Published var titles: [String]
```
JunSong-SH, line 79: Not sure why the titles, segmentIndex and segmentWidthMode were '@Published' here? From my view, I only see the '_maxSegmentWidth' is possible to be changed after the DimensionSegmentModelObject was created in the DimensionSelector. However, it was not '@Published'?

Sources/FioriSwiftUICore/_FioriStyles/DimensionSelectorStyle.fiori.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+    
+    func getHStack(_ configuration: DimensionSelectorConfiguration) -> some View {
+        if !self.isEnabled, configuration.selectedIndex != nil {
+            configuration.selectedIndex = nil
+        }
+        return HStack(alignment: .center, spacing: configuration.interItemSpacing) {
+            ForEach(configuration.titles.indices, id: \.self) { index in
+                if !configuration.segment(configuration.titles[index]).isEmpty {
```
JunSong-SH, line 34: The line 34 and 35 to call configuration.segment(configuration.titles[index]) twice. Maybe, you can simply use a variable here. For example, 
let segment = configuration.segment(configuration.titles[index])
if !segment.isEmpty {
    AnyView(segment)
}

Sources/FioriSwiftUICore/_FioriStyles/DimensionSelectorStyle.fiori.swift
```diff
@@ -0,0 +1,83 @@
...
+                    AnyView(configuration.segment(configuration.titles[index]))
+                        .environmentObject(DimensionSegmentModelObject(segmentIndex: index, _maxSegmentWidth: self._maxSegmentWidth, configuration: configuration))
```
JunSong-SH, line 36: I think you try to use 'DimensionSegmentModelObject' to pass data to DimensionSegment. However, creating an `.environmentObject` in a loop is generally not recommended. So, if you really need the DimensionSegmentModelObject, you might should init it and set it to DimensionSelector with context to avoid multi-instance and declare the _maxSegmentWidth as Published. 
For the title of the DemensionSegment, maybe, you can let DemensionSegment inherits the TitleComponent. Thus, you don't need the 'segmentIndex' and only init the DemensionSegment with 'title' and 'isSelected'. Same logic for App team for customization base on your code logic.

JunSong-SH (approved):
LGTM!

xiaoqinggrace (approved):

PR #973, opened 2025-02-07
fix: 🐛 [JIRA: IOSSDKBUG-548] BannerMultiMessageSheet optimization

hengyi-zhang:
BannerMultiMessageSheet component supports customization to hide view detail link, or close action.


dyongxu (approved):
cherrypick to rel-4.3

PR #974, opened 2025-02-07
fix: 🐛 [IOSSDKBUG-547]FilterFeedbackBar list duplicated names

restaurantt:
When options for the list have duplicated names, distinguish selected name by adding UUID as key and name as value.


dyongxu (changes requested):
Since the old style is deprecated (ModelDefinitions.swift), we don't have to update this file. Any new change should not rely on this file.

dyongxu (approved):

PR #987, opened 2025-02-20
refactor: 💡 [HCPSDKFIORIUIKIT-2888] EmptyStateView Refactor

zuopengchi:
Refactor `EmptyStateViewModel`. Since the existing IllustrationMessage view offers the same functionality as EmptyStateView, It is recommended that the developer use the `IllustratedMessageComponent` instead.


dyongxu (commented):
Can you check what kind of changes we may have to make , in SAPFiori code base?

PR #989, opened 2025-02-24
feat: 🎸 HCPSDKFIORIUIKIT-2878 Refactor KPIItemModel

zzchao-1999:
Refactor KPIItemModel:
KPIItem


Sources/FioriSwiftUICore/_ComponentProtocols/BaseComponentProtocols.swift
```diff
@@ -640,3 +640,11 @@ protocol _KeyComponent {
     // sourcery: @ViewBuilder
     var key: AttributedString { get }
 }
+
+// sourcery: BaseComponent
+protocol _KPIViewSubItemComponent {
```
xiaoyu0722, line 645: It's a sub-item for `KPIItem`, can this be renamed as `_KPISubItemComponent`?

Sources/FioriSwiftUICore/_FioriStyles/KPIItemStyle.fiori.swift
```diff
@@ -0,0 +1,106 @@
...
+            .onChanged { _ in
+                self.isPressed = true
+            }
+            .onEnded { _ in
+                self.isPressed = false
+            }
+    }
+}
+
+struct buildKPIItems: View {
```
xiaoyu0722, line 32: First character should be upper-case.
If this only work for KPIItem, we should wrap these into a function in `KPIItemStyle`. No need to make it as a global struct.

Sources/FioriSwiftUICore/_FioriStyles/KPIItemStyle.fiori.swift
```diff
@@ -0,0 +1,106 @@
...
+                    .foregroundStyle(getForegroundColor(for: getConfigState(isEnabled: self.isEnabled, isPressed: self.isPressed)))
+                    .font(getFont(forType: item.kPIViewSubItemType, typeSize: self.configuration.proposedViewSize))
+            }
+        }
+        .frame(maxWidth: .infinity, alignment: self.configuration.alignment)
+        .lineLimit(1)
+    }
+}
+
+func getFont(forType type: KPISubitemType, typeSize: KPIItemSize) -> Font {
```
xiaoyu0722, line 53: Same as above, this should not be a global function.

Sources/FioriSwiftUICore/_FioriStyles/KPIItemStyle.fiori.swift
```diff
@@ -0,0 +1,106 @@
...
+    switch type {
+    case .icon:
+        switch typeSize {
+        case .large:
+            return Font.fiori(fixedSize: 18)
+        case .small:
+            return Font.fiori(fixedSize: 13.5)
+        default:
+            return Font.fiori(fixedSize: 13.5)
```
xiaoyu0722, line 62: Is this should be a font style but not a fixed size?

Sources/FioriSwiftUICore/_FioriStyles/KPIItemStyle.fiori.swift
zzchao-1999: Updated it as func and moved it into baseStyle struct.

Sources/FioriSwiftUICore/_FioriStyles/KPIItemStyle.fiori.swift
zzchao-1999: Moved it into baseStyle struct.

Sources/FioriSwiftUICore/_FioriStyles/KPIItemStyle.fiori.swift
zzchao-1999: Updated them with font style.

PR #1012, opened 2025-03-07
fix: 🐛 remove extra vertical spacing in object item

xiaoyu0722:
Use empty view for footnote and icons. Then VStack will handle the spacing automatically. And remove unnecessary bottom padding for footnote and icons.


dyongxu (approved):

dyongxu:
This is for IOSSDKBUG-592

PR #1018, opened 2025-03-10
fix: 🐛 [JIRA:0] BannerMultiMessage enhancement

hengyi-zhang:
Workaround for forcing list refresh when second layer array modified in bannerMultiMessage.


dyongxu (changes requested):
Build failure to fix:
❌  /Users/runner/work/cloud-sdk-ios-fiori/cloud-sdk-ios-fiori/Sources/FioriSwiftUICore/_FioriStyles/BannerMultiMessageSheetStyle.fiori.swift:366:20: instance method 'appendInterpolation(_:formatter:)' requires that 'Bool' inherit from 'NSObject'
            Text("\(self.refreshFlag)")

dyongxu (approved):

PR #1019, opened 2025-03-10
refactor: 💡 [HCPSDKFIORIUIKIT-2882] WelcomeScreen refactor

hengyi-zhang:


Sources/FioriSwiftUICore/Models/ModelDefinitions.swift
```diff
@@ -202,6 +202,10 @@ public protocol WelcomeScreenModel: TitleComponent, DescriptionTextComponent, Su
     var secondaryAction: _ActionModel? { get }
 }
 
+/// Deprecated WelcomeScreenModel
+@available(*, unavailable, renamed: "_WelcomeScreenModel", message: "Will be removed in the future release. Please create WelcomeScreen with other initializers instead.")
```
dyongxu, line 206:
```suggestion
@available(*, deprecated, renamed: "_WelcomeScreenModel", message: "Will be removed in the future release. Please create WelcomeScreen with other initializers instead.")
```

dyongxu:
Screenshot 2025-03-10 at 3 31 11 PM
The left side is SwiftUI , right side is UIKit.  
1. Spacing is not the same 
2. The text color for "Try the demo" is not the same
@hengyi-zhang

dyongxu (changes requested):
comments inlined

Sources/FioriSwiftUICore/Models/ModelDefinitions.swift
hengyi-zhang: Under `@available(*, unavailable, renamed: "_WelcomeScreenModel", message: "Will be removed in the future release. Please create WelcomeScreen with other initializers instead.")`, the protocol is empty. I think `deprecated` means it can be used in new version too.

dyongxu: yes, that is correct. Let's use unavailable here.

dyongxu (approved):

PR #1030, opened 2025-03-27
chore: 🤖 fix localization string

dyongxu:


dyongxu:
Merged without explicit approval, as it blocks translation process

PR #1035, opened 2025-04-07
fix: 🐛 build failed on visionOS

dyongxu:
1. Add conditional compiler flag to visionOS and iOS
2. Add MandatoryAttachmentExample.swift into project, which was missing earlier and causing build failure as well
3. For **'camera' is unavailable in visionOS** , need further code investigation and clean up to remove camera API in other places where it should be used.  (e.g. in the menu to choose attachment)


CharlesXu0488 (approved):
will take care camera in next pr completely.

PR #1051, opened 2025-04-22
fix: 🐛 [JIRA:0] Fixed the text input issue for Stepper

xiaoqinggrace:
Added decimal style for TextInputField


Sources/FioriSwiftUICore/_FioriStyles/TextInputFieldStyle.fiori.swift
```diff
@@ -38,28 +38,65 @@ extension TextInputFieldFioriStyle {
 }
 
 /// Number style
```
dyongxu, line 40:
```suggestion
/// Decimal style
```

dyongxu (changes requested):
comment to be updated

PR #1055, opened 2025-04-24
fix: 🐛 Enhance stepper validation

zzchao-1999:


Sources/FioriSwiftUICore/_FioriStyles/TextInputFieldStyle.fiori.swift
xiaoqinggrace, line 1: We should not remove the validation for TextInputField styling since applications in different use cases reuse it.

billzhou0223 (approved):

dyongxu (approved):
Pass this review so it can be merged. But consider to add more unit test when bandwidth allows.

PR #1057, opened 2025-04-24
fix: 🐛 IOSSDKBUG-688 invalid format of PrivacyInfo.xcprivacy

dyongxu:


dyongxu:
Did not find the official word from Apple website. But Chatgpt says:
**Issues in Your File
Empty Strings:
<string></string> — Apple doesn’t allow blank string values. Every required key must have a valid, documented value.
Empty dict in NSPrivacyAccessedAPITypes:
<dict/> — A dictionary here must include specific keys like NSPrivacyAccessedAPIType, NSPrivacyAccessedAPITypeReasons, etc.**

So, clean up the file to remove the empty "string and dict".
Also remove all entries under NSPrivacyAccessedAPITypes since there is no access required there.

Summary:  PrivacyInfo.xcprivacy is updated to the simplest form , requiring no privacy access / collection.

dyongxu:
Merge this for now to build a milestone drop so that Sam can do a new TestFlight upload to verify.

PR #1069, opened 2025-05-06
fix: 🐛 [IOSSDKBUG-708] BannerMultiMessage enhancement

hengyi-zhang:
Support to hide close action、swipe delete action and clear action under category. Add simultaneous tap gesture on the top-right close action to dismiss the sheet.

Modify close action default image.


Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -679,7 +679,12 @@ protocol _ToastMessageComponent: _IconComponent, _TitleComponent {
 }
 
 // sourcery: CompositeComponent
-protocol _BannerMultiMessageSheet: _TitleComponent, _CloseActionComponent {
+// sourcery: importFrameworks = ["FioriThemeManager"]
+protocol _BannerMultiMessageSheet: _TitleComponent {
+    @ViewBuilder
+    // sourcery: defaultValue = "{ FioriIcon.status.error }"
+    // sourcery: resultBuilder.defaultValue = "{ FioriIcon.status.error }"
+    var closeAction: (() -> any View)? { get }
```
billzhou0223, line 687: Why make this change? It should be inherited from `_CloseActionComponent`.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
hengyi-zhang: BannerMultiMessageSheet's close action has different default icon. If it's inherited from _CloseActionComponent, there is no way to customize my own default icon in init api parameters in generated file. So I move the property into the body and not inherited from another protocol.

billzhou0223 (approved):

xiaoqinggrace (approved):

PR #1090, opened 2025-05-26
feat: 🎸 [HCPSDKFIORIUIKIT-2951]TextFields support Currency

KevinZK:
* Fix the click area issue of the placeholderTextfield component cancel button. 
  This results in the entire textfieldformView being able to respond to the button's click event even when the input box is activated

Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-05-28 at 16 32 01

### DIFF
--- a/Sources/FioriSwiftUICore/Utils/Currency.swift
+++ b/Sources/FioriSwiftUICore/Utils/Currency.swift
@@ -1,41 +1,99 @@
 import Foundation
 import SwiftUI
 
+/// An environment key for configuring currency field behavior in SwiftUI views.
 struct CurrencyFieldKey: EnvironmentKey {
-    static let defaultValue = CurrencyFieldConfiguration(isCurrencyField: false, formatter: NumberFormatter.currencyDefault)
+    /// The default value for the currency field configuration.
+    static let defaultValue = CurrencyFieldConfiguration(isCurrencyField: false, formatter: .numberFormat(NumberFormatter.currencyDefault))
 }
 
 extension EnvironmentValues {
+    /// The currency field configuration for the current environment.
     var currencyField: CurrencyFieldConfiguration {
         get { self[CurrencyFieldKey.self] }
         set { self[CurrencyFieldKey.self] = newValue }
     }
 }
 
+/// Configuration for a currency input field, including whether it is a currency field and the formatter to use.
 struct CurrencyFieldConfiguration {
+    /// Indicates if the field is a currency input field.
     let isCurrencyField: Bool
-    let formatter: NumberFormatter
+    /// The formatter used for currency formatting.
+    let formatter: formatType
+}
+
+/// Protocol for currency formatting, providing locale and formatting functionality.
+public protocol currencyFormatting {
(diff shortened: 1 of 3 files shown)


dyongxu (changes requested):
For the first testcase, which is "price(USD)*":
1. Add a switch to turn on/off the mandatory field
2. Add a picker to allow the user to choose one of the available locales  to decide the ISO code, so that "USD" can show as different ISO code

For the validation message, add some real validation logic. e.g check if the enter value is larger than 5.0.

Investigate why the entered value disappears if the user clicks outside the input field. This will cause the lost of user input.

For "max Text Length:15" testcase, need to allow decimal input, to show the "." is counted as part of the 15 max.

KevinZK:
> For the first testcase, which is "price(USD)*":
> 
> 1. Add a switch to turn on/off the mandatory field
> 2. Add a picker to allow the user to choose one of the available locales  to decide the ISO code, so that "USD" can show as different ISO code
> 
> For the validation message, add some real validation logic. e.g check if the enter value is larger than 5.0.
> 
> Investigate why the entered value disappears if the user clicks outside the input field. This will cause the lost of user input.
> 
> For "max Text Length:15" testcase, need to allow decimal input, to show the "." is counted as part of the 15 max.

Done

Sources/FioriSwiftUICore/Utils/Currency.swift
```diff
@@ -0,0 +1,82 @@
...
+    var currencyField: CurrencyFieldConfiguration {
+        get { self[CurrencyFieldKey.self] }
+        set { self[CurrencyFieldKey.self] = newValue }
+    }
+}
+
+struct CurrencyFieldConfiguration {
+    let isCurrencyField: Bool
+    let formatter: NumberFormatter
+}
```
agrabz, line 18: @dyongxu offered me to review this PR and overall it looks really good. Since we requested this feature we found a much better way for the formatting which is below.

Instead of NumberFormatter I suggest to use the below API that can handle:
1. ISO code transformation to currency symbol (if any)
2. symbol/ISO Code placement (before or after the value)
3. grouping and decimal separator based on the locale.

Additionally it is based on Decimal so customers cannot use Double, Float or Int accidentally, which would not work properly.

This is an extract from our source code:
```swift
Decimal.FormatStyle.Currency(
            code: self.currency ?? "",
            locale: locale
        )
        .decimalSeparator(strategy: .always)
        .grouping(.automatic)
        .precision(.fractionLength(fractionalDigits))
        .presentation(.standard)
        .format(value)
```

The full context is here: https://github.tools.sap/One-Mobile-Experience/sarm-ios/blob/main/src/SARMCommonUtils/Sources/SARMCommonUtils/Models/Money.swift

KevinZK: When combined with SwiftUI, cross-platform, and declarative UI applications, the FormatStyle API is more concise and efficient
While NumberFormatter is more suitable for maintaining old projects or scenarios where it is used in combination with UIKit.

I will add an implementation of FormatStyle.

PR #1096, opened 2025-06-06
refactor: 💡 introspect upgrade

xiaoyu0722:
Introspection upgrade.
Find the right view in multiple same type views.

### Important: MIT License Review Required
Source: https://github.com/siteline/swiftui-introspect


dyongxu (changes requested):
We shall not add source code directly into the repro.
Instead, we shall update Package.swift to add the dependency.

let package = Package(
    dependencies: [
        .package(url: "https://github.com/siteline/swiftui-introspect", from: "1.0.0"),
    ],
    targets: [
        .target(name: <#Target Name#>, dependencies: [
            .product(name: "SwiftUIIntrospect", package: "swiftui-introspect"),
        ]),
    ]
)

PR #1102, opened 2025-06-13
refactor: 💡 import introspection

xiaoyu0722:
import [introspection](https://github.com/siteline/swiftui-introspect.git)


Package.swift
```diff
@@ -25,6 +25,9 @@ let package = Package(
             targets: ["FioriThemeManager"]
         )
     ],
+    dependencies: [
+        .package(url: "https://github.com/siteline/swiftui-introspect.git", .upToNextMajor(from: "1.0.0"))
```
dyongxu: Suggest to change it to 
.exact("1.3.0")
So, it will not automatically pick up patch releases, which we may not need.

dyongxu (changes requested):
update to exact version

billzhou0223 (approved):

dyongxu (approved):

PR #1111, opened 2025-06-17
fix: 🐛 extra top spacing for description in object item

xiaoyu0722:
3 vertical spacing for title, subtitle..., which is right in code.
Just extra top spacing needed for description.
I can't find a latest spec for specific spacing between each elements. I'm not sure if this one is out of date.

|spec|app view|
|-|-|
|Screenshot 2025-06-17 at 15 57 37|Simulator Screenshot - iPhone 16 Pro Max - 2025-06-17 at 16 15 12|


dyongxu:
@xiaoyu0722 need cherrypick to rel-25.4

PR #1117, opened 2025-06-18
feat: 🎸 [HCPSDKFIORIUIKIT-2852] writing assistant

xiaoyu0722:


dyongxu (commented):
1. Can we have some example to see how WA can replace only the selected text ? Other text remains unchanged. 
2. The error view layout for "Make Bullet List" is not correct.
3. WACommand might be better to be called "WAMenu".
4. When using the sub menu, e.g. "Change Tone",  is there a need to have a way to go back the main menu? (I have not verified with the spec. So, it is a question for now)

PR #1123, opened 2025-06-23
feat: 🎸 [FIORIDESIGN-14795]Demo for card add left image

KevinZK:
Demo

### Scope
1. Card media should be a square 1:1 image
2. Card content (header + body) should be to the right of the image
3. Card footers are not used in our cases for these cards, but can potentially also be fit in the container to the right of the image
4. Image should be edge-to-edge as shown in the example above


dyongxu:
@KevinZK Some early comment:
The "Left MediaImage" should be not a new component.  It is the existing MediaImage, but to be layout on the left side

KevinZK:
> @KevinZK Some early comment: The "Left MediaImage" should be not a new component. It is the existing MediaImage, but to be layout on the left side

Got it, Will solve

PR #1138, opened 2025-07-01
fix: 🐛 [IOSSDKBUG-774] List row background in OrderPicker

zuopengchi:
1. Set the default background of the list row to be transparent
2. Add a property to the OrderPickerItemModel to allow setting a custom background color for the list row.

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/OrderPickerStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/OrderPickerStyle.fiori.swift
@@ -22,6 +22,7 @@ public struct OrderPickerBaseStyle: OrderPickerStyle {
                     .accessibilityElement(children: .combine)
                     .accessibilityLabel(item.wrappedValue.description)
                     .accessibilityAddTraits(.isButton)
+                    .listRowBackground(item.wrappedValue.customListRowBackground ?? Color.clear)
                 }.onMove(perform: self.moveItems)
             }
         }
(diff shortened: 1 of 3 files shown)


Sources/FioriSwiftUICore/_FioriStyles/OrderPickerStyle.fiori.swift
```diff
@@ -112,6 +113,8 @@ public struct OrderPickerItemModel: Identifiable, Hashable, Equatable, CustomStr
     
     /// An optional `customStyle`
     public var customStyle: (any SortCriterionStyle)?
+    /// An optional color that determins the background color of the custom list row.
```
dyongxu:
```suggestion
    /// An optional color that determines the background color of the custom list row.
```

dyongxu (changes requested):
API doc change

Sources/FioriSwiftUICore/_FioriStyles/OrderPickerStyle.fiori.swift
zuopengchi: Changed

dyongxu:
@zuopengchi Need cherrypick to rel-25.4

dyongxu (approved):

zuopengchi:
> @zuopengchi Need cherrypick to rel-25.4

The Order Picker is a new feature that is not included in the rel-25.4 release

PR #1140, opened 2025-07-01
fix: 🐛 [IOSSDKBUG-783]SwiftUI Example can't compile for Mac Catalyst

zzchao-1999:


dyongxu:
@zzchao-1999 Can the filter of Framework be included as well, as part of the Example.xcodeproj ?

zzchao-1999:
> @zzchao-1999 Can the filter of Framework be included as well, as part of the Example.xcodeproj ?

Mac Catalyst is not enabled by default, so the platform filter is unavailable in the embedded content. We only need to add the filter to Example.xcodeproj if Mac Catalyst is included in the default destination list.

dyongxu (approved):

PR #1156, opened 2025-07-11
fix: 🐛 writing assistant keyboard input failure

xiaoyu0722:
Keyboard input failure when W.A. modifier added.


Sources/FioriSwiftUICore/Utils/SkeletonLoading.swift
```diff
@@ -247,17 +266,21 @@ public struct SkeletonLoadingContainer<Content: View>: View {
...
-                    .redacted(reason: .placeholder)
-                    .skeletonLoading(isTintColor: self.isTintColor)
-            } else {
-                self.content()
-            }
+            self.content()
+                .foregroundColor(self.redactedForegroundColor)
+                .redacted(reason: self.isLoading ? .placeholder : [])
+                .skeletonLoading(isLoading: self.isLoading, isTintColor: self.isTintColor)
         }
```
xiaoqinggrace: The skeleton loading will not be applied when isLoading is false based on the original implementation. There is no effect on the performance. With the change in the PR, all the skeleton loading modifier will be checked and applied. It would be better to avoid since  the skeleton loading is only shown in short time.

xiaoyu0722: Hi @xiaoqinggrace , I can see what you worried is you don't want to apply unnecessary modifiers when the view is not loading.
The key problem, when using if else will cause a new View created.
Differences like:
```swift
if isLoading {
    Text("a")
} else {
    Text("a").foregroundColor(Color.red)
}

Text("a").foregroundColor(isLoading ? Color.clear : Color.red)
```
The first one will create different views when `isLoading` changed, then wa will create a different context.
But the second one will keep the view consistent.

Like W.A. can only worked like
```swift
// This is what really needed.
TextEditor.modifier(isLoading).wa...
```

But not
```swift
// Actually this will cause two different flows when isLoading changed.
if isLoading {
    TextEditor().wa...
} else {
    LoadingTextEditor().wa...
}

Group {
    if isLoading {
        TextEditor()
    } else {
        LoadingTextEditor()
    }
}.wa....
```

xiaoyu0722: Also if you could check the logic and see if the modifier applied may cause unexpected fixed effect.
And no need worry about the memory overhead, I remember `View` is very cheap in SwiftUI.

If you really want the previous implementation back, I will continue to try if there is a better way to fix W.A. loading issue.

xiaoqinggrace (approved):

PR #1157, opened 2025-07-11
fix: 🐛 Fix incorrect formatting of binding value in getter

KevinZK:
### Description:
In the input box that supports currency format.
`text: Binding<String>` property: The text used by developers should not be formatted and should read the original input value.

### Solve：
Provide properties for display only`@State private var displayText: String = ""`

### Reference：
The correct usage logic should be consistent with Apple's API.
Apple API：`nonisolated public init<V>(_ titleKey: LocalizedStringKey, value: Binding<V>, formatter: Formatter)`
`TextField(<#T##LocalizedStringKey#>, value: <#T##Binding<V>#>, formatter: <#T##Formatter#>)`

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/TextFieldFormViewStyle.fiori.swift
@@ -62,8 +62,8 @@ extension TextFieldFormViewFioriStyle {
         @Environment(\.currencyField) private var currencyFieldConfiguration
         @FocusState var isFocused: Bool
         @State var isEditing: Bool = false
-        @State private var formattedText: String = ""
         @State private var rawInput: String = ""
+        @State private var displayText: String = ""
         @Environment(\.isLoading) var isLoading
         @Environment(\.isAILoading) var isAILoading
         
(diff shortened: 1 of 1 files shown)


KevinZK:
> Can you check , why non digit characters are allowed for currency field ? I think we should only allow digit or "." for this field.

I set the type of the keyboard to be a `.decimalPad`.
The keyboard can be switched back to the characters keyboard.
If we want to completely filter out characters and special symbols, we need to apply forced regularization to the input data.

I just applied regular expressions to the input data during formatting.

This logic is also consistent with `Apple API: nonisolated public init<V>(_ titleKey: LocalizedStringKey, value: Binding<V>, formatter: Formatter)`

dyongxu (changes requested):
I saw the the overlap on top of the placeholder text is fixed.
However, I still see the input allows the non-digital characters.

janhuachu:
> I saw the the overlap on top of the placeholder text is fixed. However, I still see the input allows the non-digital characters. I think we should not allow these non-digital characters for input .

I think we should port the FUINumberFormatter to SwiftUI side and allows the SwiftUI implementation of TextFieldFormView to accept a NumberFormatter. This should be part of this feature request https://jira.tools.sap/browse/HCPSDKFIORIUIKIT-2957

KevinZK:
> > I saw the the overlap on top of the placeholder text is fixed. However, I still see the input allows the non-digital characters. I think we should not allow these non-digital characters for input .
> 
> I think we should port the FUINumberFormatter to SwiftUI side and allows the SwiftUI implementation of TextFieldFormView to accept a NumberFormatter. This should be part of this feature request https://jira.tools.sap/browse/HCPSDKFIORIUIKIT-2957

The current API supports the passing of NumberFormatter.
`.currencyField(formatter: .numberFormat(NumberFormatter(localeIdentifier: "de_DE", maxFractionDigits: 2)))`

KevinZK:
> I saw the the overlap on top of the placeholder text is fixed. However, I still see the input allows the non-digital characters. I think we should not allow these non-digital characters for input .

Done

PR #1181, opened 2025-07-25
fix: 🐛 [IOSSDKBUG-845] ContactItem improvement

hengyi-zhang:
Modify the logic when using the splitPercent property.


shengxu7 (changes requested):
if self.horizontalSizeClass == .some(.regular) {
    self.bodyInRegular(configuration)
} else {
    self.bodyInCompact(configuration)
}, this code does not work if two contact items are shown side by side in an iPad. self.horizontalSizeClass is .regular but actual the view should be treated as .compact.

hengyi-zhang:
> if self.horizontalSizeClass == .some(.regular) { self.bodyInRegular(configuration) } else { self.bodyInCompact(configuration) }, this code does not work if two contact items are shown side by side in an iPad. self.horizontalSizeClass is .regular but actual the view should be treated as .compact.

Fixed!

Sources/FioriSwiftUICore/_FioriStyles/ContactItemStyle.fiori.swift
```diff
@@ -9,12 +9,16 @@ public struct ContactItemBaseStyle: ContactItemStyle {
...
     
+    @State var compactBodyWidth: CGFloat = 0
+    
     public func makeBody(_ configuration: ContactItemConfiguration) -> some View {
         Group {
-            if self.horizontalSizeClass == .some(.regular) {
-                self.bodyInRegular(configuration)
-            } else {
+            if self.horizontalSizeClass == .some(.compact)
+                || self.compactBodyWidth <= 375.0
```
shengxu7: 375 is way too small. I used 666 in CardFooter and that number works greatly.

shengxu7 (changes requested):

dyongxu (approved):

shengxu7 (approved):

PR #1207, opened 2025-08-07
feat: 🎸 add isInMenu environment key and a misc test

shengxu7:
Tests on iOS 17.5
iOS-17 5

Tests on iOS 18.3.1
iOS-18 3 1

### DIFF
--- a/Sources/FioriSwiftUICore/Utils/EnvironmentValues.swift
+++ b/Sources/FioriSwiftUICore/Utils/EnvironmentValues.swift
@@ -21,7 +21,7 @@ public extension EnvironmentValues {
     }
 }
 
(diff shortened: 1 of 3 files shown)


Sources/FioriSwiftUICore/Utils/EnvironmentValues.swift
```diff
@@ -20,3 +20,16 @@ public extension EnvironmentValues {
         }
     }
 }
+
+/// The key for storing whether the component is in a `Menu` in the environment.
```
dyongxu:
```suggestion
/// The key for storing whether the component is in a `Menu` in the environment. Default value is false.
```

dyongxu (approved):

Issue #1162, opened 2025-07-15
TextFieldFormView no support for Formatter


agiguere:
The struct TextFieldFormView has no support yet for formatter, the equivalent API in UIKit does have, this is something required to format currency or number

thanks
alex

leoxiao-sap:
Hi @agiguere,
We already support phone number, currency, number, and generic text formatting for TextFieldFormView. You can see the examples in TextFieldFormView.swift. Thanks!
cc: @KevinZK

PR #1236, opened 2025-08-26
feat: 🎸 [HCPSDKFIORIUIKIT-2957]Implement a generic AlphaNumeric

leoxiao-sap:
Dev result
Simulator Screenshot - iPhone 16 - 2025-08-25 at 18 14 29

https://github.com/user-attachments/assets/45bbd898-8ec7-4154-9b8e-1b4b78f8c4fe

### DIFF
--- a/Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
+++ b/Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
@@ -78,6 +78,8 @@ protocol _TextInputFieldComponent {
     var text: String { get set }
     // sourcery: defaultValue = false
     var isSecureEnabled: Bool? { get set }
+    // sourcery: default.value = nil
+    var formatter: GenericTextFormatter? { get set }
 }
 
 // sourcery: CompositeComponent
--- a/Apps/Examples/Examples.xcodeproj/project.pbxproj
+++ b/Apps/Examples/Examples.xcodeproj/project.pbxproj
@@ -45,6 +45,7 @@
 		64905D092C7693970062AAD4 /* DateTimePickerExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 64905D082C7693970062AAD4 /* DateTimePickerExample.swift */; };
 		649C7F482CE531C8007B78A7 /* ProgressIndicatorExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 649C7F472CE531C5007B78A7 /* ProgressIndicatorExample.swift */; };
 		64A3FF012D12527F00992B24 /* DimensionSelectorExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 64A3FF002D12527B00992B24 /* DimensionSelectorExample.swift */; };
+		684E21D32E4D7921001823E2 /* TextInputExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 684E21D22E4D7921001823E2 /* TextInputExample.swift */; };
 		691DE21925F2A30B00094D4A /* KPIItemExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 691DE21825F2A30B00094D4A /* KPIItemExample.swift */; };
 		692F338B26556A6A009B98DA /* SideBarExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 692F338A26556A6A009B98DA /* SideBarExample.swift */; };
 		69B2B5D9268A333C009AC6B3 /* KPIProgressViewExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 69B2B5D8268A333C009AC6B3 /* KPIProgressViewExample.swift */; };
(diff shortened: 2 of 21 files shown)


KevinZK:
Issue: #1162 

Should compatibility with the system API `Formatter` and `FormatStyle` be considered?
`TextField(<#T##LocalizedStringKey#>, value: <#T##Binding<V>#>, formatter: <#T##Formatter#>)`
`TextField(<#T##LocalizedStringKey#>, value: <#T##Binding<F.FormatInput?>#>, format: <#T##F#>)`

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -78,6 +78,8 @@ protocol _TextInputFieldComponent {
     var text: String { get set }
     // sourcery: defaultValue = false
     var isSecureEnabled: Bool? { get set }
+    // sourcery: default.value = nil
+    var formatter: GenericTextFormatter? { get set }
```
janhuachu: I think it is better to use FormattedStringEditing as the formatter type to allow PhoneNumberFormatter and CustomNumberFormatter.

Sources/FioriSwiftUICore/Utils/Formatter/CustomNumberFormatter.swift
```diff
@@ -0,0 +1,290 @@
...
+        func getPrefixAndSuffix(_ formattedString: String) -> (prefix: String, suffix: String) {
+            var isPositive = true
+            if let number = defaultFormatter.number(from: self.effectiveString(formattedString)) {
+                if number.doubleValue < 0 {
+                    isPositive = false
+                }
+            }
+
+            let prefix = (isPositive ? self.positivePrefix : self.negativePrefix)!
+            let suffix = (isPositive ? self.positiveSuffix : self.negativeSuffix)!
```
janhuachu: Please fix all the "Force Unwrapping Violation" warnings.

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
leoxiao-sap: Hi @janhuachu,
Based on the initial discussion with David, we believed that the Generic Text Formatter could cover the functionalities of both PhoneNumberFormatter and CustomNumberFormatter, which is why I didn’t migrate PhoneNumberFormatter and CustomNumberFormatter initially. Now, it appears that these are needed for UIKit compatibility, so I’ve renamed and added them directly. Please correct me if my understanding is inaccurate.

To support PhoneNumberFormatter and CustomNumberFormatter in SwiftUI, implementation adjustments are required because SwiftUI’s TextField doesn’t natively support cursor manipulation. Currently, in SwiftUI, the Generic Text Formatter deletes characters from the middle by jumping to the end and deleting backward. If cursor positioning behavior (like in UIKit) is needed, additional extensions would be necessary, as both PhoneNumberFormatter and CustomNumberFormatter involve cursor operations during deletion.

Moreover, the timing of inserting spaces, hyphens, and other formatting characters differs between PhoneNumberFormatter/CustomNumberFormatter and the Generic Text Formatter.

I have fixed all "Force Unwrapping Violation" warning.
Thank you
cc: @dyongxu

leoxiao-sap: Hi @janhuachu,
Based on the initial discussion with David, we believed that the Generic Text Formatter could cover the functionalities of both PhoneNumberFormatter and CustomNumberFormatter, which is why I didn’t migrate PhoneNumberFormatter and CustomNumberFormatter initially. Now, it appears that these are needed for UIKit compatibility, so I’ve renamed and added them directly. Please correct me if my understanding is inaccurate.

To support PhoneNumberFormatter and CustomNumberFormatter in SwiftUI, implementation adjustments are required because SwiftUI’s TextField doesn’t natively support cursor manipulation. Currently, in SwiftUI, the Generic Text Formatter deletes characters from the middle by jumping to the end and deleting backward. If cursor positioning behavior (like in UIKit) is needed, additional extensions would be necessary, as both PhoneNumberFormatter and CustomNumberFormatter involve cursor operations during deletion.

Moreover, the timing of inserting spaces, hyphens, and other formatting characters differs between PhoneNumberFormatter/CustomNumberFormatter and the Generic Text Formatter.

For "Force Unwrapping Violation" warning, I’ll attempt to fix them.
Thank you
cc: @dyongxu

janhuachu: @leoxiao-sap I think it does not make sense if you just move FUINumberFormatter and FUIPhoneNumberFormatter to SwiftUI part without changing the formatter type to optional GenericTextFormatter.
I think it is better to make the formatter type to FormattedStringEdting and move the 2 formatters to the SwiftUI side.

leoxiao-sap: @janhuachu 
Thank you for your answer, I will do this update.

leoxiao-sap: > @leoxiao-sap I think it does not make sense if you just move FUINumberFormatter and FUIPhoneNumberFormatter to SwiftUI part without changing the formatter type to optional GenericTextFormatter. I think it is better to make the formatter type to FormattedStringEdting and move the 2 formatters to the SwiftUI side.

@janhuachu I have completed this update. Could you please review it? Thank you!

janhuachu (approved):
LGTM

PR #1252, opened 2025-09-05
chore: 🤖 fix fatalError when running tests of a package that relies on FioriSwiftUI package

MarcoEidinger:
Fix https://jira.tools.sap/browse/IOSSDKBUG-929

The issue was previously fixed with https://github.com/SAP/cloud-sdk-ios-fiori/pull/728 last year, but the relevant code seems to have been deleted in April this year with https://github.com/SAP/cloud-sdk-ios-fiori/pull/1041

### DIFF
--- a/Sources/FioriSwiftUICore/Utils/Bundle+Extensions.swift
+++ b/Sources/FioriSwiftUICore/Utils/Bundle+Extensions.swift
@@ -5,6 +5,17 @@ private class CurrentBundleFinder {}
 extension Bundle {
     static var accessor: Bundle {
         #if SWIFT_PACKAGE
+
+            // patch for SwiftPM: ensure that bundle is found when accessed from a package that relies on FioriSwiftUI
+            if ProcessInfo.processInfo.processName == "xctest" {
+                // two names are evaluated because this source file is shared between FioriThemeManager & FioriSwiftUICore modules
+                for bundleName in ["FioriSwiftUI_FioriThemeManager", "FioriSwiftUI_FioriSwiftUICore"] {
+                    if let bundle = patchToFindBundle(with: bundleName) {
+                        return bundle
+                    }
+                }
+            }
+        
             return Bundle.module
         #else
             return Bundle(for: CurrentBundleFinder.self)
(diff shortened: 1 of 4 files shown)


MarcoEidinger:
@dyongxu I did not test it myself. I asked Akos to verify. I have to sign off for today so please take over. Thanks!

MarcoEidinger:
Screenshot 2025-09-05 at 6 50 42 PM

MarcoEidinger:
But now some of our own tests are not working

Screenshot 2025-09-05 at 6 51 59 PM

I can look into that on Tuesday. Until then, I recommend to hold off with a merge.

MarcoEidinger:
But only the test is really affected and not productive code, so the change could be shipped and the test could be fixed later

dyongxu:
> But now some of our own tests are not working
> 
> Screenshot 2025-09-05 at 6 51 59 PM
> I can look into that on Tuesday. Until then, I recommend to hold off with a merge.

I have tried this PR locally. But did not see the failure. Did not find why it did not repro locally yet.

dyongxu:
@MarcoEidinger Since this only affects the test , I decided to do the 25.8.1 patch first, without this PR.

dyongxu (changes requested):
In Locale+Extension.swift, Bundle.accessor is resolved as "FioriThemingManager", instead of "FioriSwiftUICore". It caused the test failure since it did not load up the correct localization file (de.lproj).

I have added a specific accessor as Bundle extension:
extension Bundle {
    /// Accessor to the resource bundle of FioriSwiftUICore.
    static var fioriCoreAccessor: Bundle {
        // resolves to the one from FioriSwiftUICore
        let bundleName = "FioriSwiftUI_FioriSwiftUICore"
        if let bundle = patchToFindBundle(with: bundleName) {
            return bundle
        }
        return Bundle.module
    }
}

Then use it in Locale+Extension.swift:
let path = Bundle.fioriCoreAccessor.path(forResource: resourceIdentifier, ofType: "lproj")

This will pass the tests

dyongxu:
@MarcoEidinger Need a cherrypick to main

PR #1270, opened 2025-09-19
fix: 🐛 [HCPSDKFIORIUIKIT-3058] toolbar layout calculation error

xiaoyu0722:
Toolbar layout updated for iOS 26, needs more padding and also bug fix for toolbar with helper text

Ref: #1267


dyongxu (changes requested):
Fix build/test failure

PR #1271, opened 2025-09-20
fix: 🐛 Restore original illustrations and add v3 illustrations

JuliaSchnSAP:


KevinZK:
David said：as per our discussion with design, the plan is to publish it into 25.11. But not planned for 25.8 as a patch.

dyongxu (commented):
Do not merge yet. This is not supposed to cherrypick into rel-25.8. Under discussion for agreement.

PR #1282, opened 2025-10-02
feat: 🎸 [HCPSDKFIORIUIKIT-3053] App level Locale override

janhuachu:
New class FioriLocale for locale override in SwiftUI part

✅ Closes: HCPSDKFIORIUIKIT-3053


Sources/FioriSwiftUICore/Utils/FioriLocale.swift
```diff
@@ -0,0 +1,56 @@
+import Foundation
+import SwiftUI
+
+/**
+ A singleton object to customize the `Locale` to be used.
+ */
+public class FioriLocale: NSObject {
+    /**
+     The shared signleton `FioriLocale`.
```
billzhou0223:
```suggestion
     The shared singleton `FioriLocale`.
```

Sources/FioriSwiftUICore/Utils/FioriLocale.swift
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public static let shared = FioriLocale()
+
+    /**
+     The customized `Locale` to be used to display the default messages by SAPFiori components.
+
+     The default valie is `nil`, which means using the device locale setting.
```
billzhou0223:
```suggestion
     The default value is `nil`, which means using the device locale setting.
```

Sources/FioriSwiftUICore/Utils/FioriLocale.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+public extension View {
+    /**
+     Sets the `locale` of the singleton `FioriLocale`.
+     - parameter locale: The desired `Locale` to be used.
+     - returns: The `View` itself.
+     */
+    func setFioriLocale(_ locale: Locale?) -> some View {
```
billzhou0223, line 67: Can you explain why this API is needed?

If the app developer wants to override the locale, they can just set the environment by themselves in app right?

```
DateTimePicker()
    .environment(\.locale, myLocale)
```

janhuachu: Please refer to this code in ContentView.swift line 48:
`.setFioriLocale(self.fioriLocale)`
It is to set the FioriLocale.shared.locale as a view modifier. I just think this is convenient for the app to use. Otherwise, the app just need to write codes to set FioriLocale.shared.locale.

Our components are not handling the ".environment(\.locale, locale)" yet. When you set the locale env, the only texts changes are the iOS DatePicker in our DateTimePicker and DateRangePicker.

It will take a lot of efforts to make the use of \.locale environment work for our components. I have discussed with David and make the decision to use this Global first and adding the support for \.locale environment later in our components.

Sources/FioriSwiftUICore/Utils/FioriLocale.swift
janhuachu: Fixed in the 2nd commit

Sources/FioriSwiftUICore/Utils/FioriLocale.swift
janhuachu: Fixed in the 2nd commit.

billzhou0223 (approved):

PR #1299, opened 2025-10-19
test: 💍 add unit test for FioriButton

dyongxu:
Screenshot 2025-10-19 at 3 29 20 PM

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/CardFooterStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/CardFooterStyle.fiori.swift
@@ -197,7 +197,7 @@ private struct CardFooterLayout: Layout {
         
         let y = maxHeight / 2
         // Move the hidden buttons out of visible area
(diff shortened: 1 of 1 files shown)


Tests/FioriSwiftUITests/FioriSwiftUICore/FioriButtonStyleTests.swift
```diff
@@ -0,0 +1,169 @@
...
+    func testContainerViewPositions() {
+        let positions: [FioriButtonImagePosition] = [.top, .leading, .bottom, .trailing]
+        for pos in positions {
+            let cfg = self.makeConfig(state: .normal, imagePosition: pos)
+            let view = cfg.containerView(.unspecified)
+            // Just ensure view builds for each position
+            _ = Mirror(reflecting: view)
+        }
+    }
+
```
hengyi-zhang, line 40: I don't think those test methods without equal compare asserts make any sense.

PR #1303, opened 2025-10-22
fix: 🐛 [IOSSDKBUG-991] revert segment control style in swift5

xiaoyu0722:


Sources/FioriSwiftUICore/_FioriStyles/SegmentedControlPickerStyle.fiori.swift
```diff
@@ -52,11 +52,15 @@ extension SegmentedControlPickerFioriStyle {
         }
         
         var borderShape: some Shape {
-            if #available(iOS 26.0, *) {
-                return Capsule()
-            } else {
+            #if swift(>=6.0)
```
dyongxu: Our SDK is using Swift compiler 5.9.  (See Package.swift, the first line)
Mobile Start team consumes it as binary framework(not source code)

That means, our binary framework will always return RoundedRectangle(cornerRadius: 9) at line 62.  No capsule will be returned at all.
For Mobile Start team, even if they use Swift compiler 6.0, they will not reach the "if" side of the code.

dyongxu (changes requested):

PR #1304, opened 2025-10-22
fix: 🐛 [IOSSDKBUG-991] revert segment control style in swift5

xiaoyu0722:


dyongxu (changes requested):
same comment in https://github.com/SAP/cloud-sdk-ios-fiori/pull/1303/files

dyongxu (approved):

PR #1306, opened 2025-10-23
feat: 🎸 [HCPSDKFIORIUIKIT-3063]Semantic Message Alignment

KevinZK:
Evaluate semantic message issues (particuarly with Text Input components: text field form cell, key value form cell, note form cell, title form cell) to align with Android:
1.character count 
      .Padding on right side
      .error state color change
2.typing state Label color 
3.Icon inside text box: alignment with text & behavior with expanding text box 
4.error message prioritization over other semantic messages


KevinZK:
close this, new #1308

PR #1308, opened 2025-10-24
feat: 🎸 [HCPSDKFIORIUIKIT-3063]Semantic Message Alignment

KevinZK:
Evaluate semantic message issues (particuarly with Text Input components: text field form cell, key value form cell, note form cell, title form cell) to align with Android:
1.character count
.Padding on right side
.error state color change
2.typing state Label color
3.Icon inside text box: alignment with text & behavior with expanding text box
4.error message prioritization over other semantic messages

### DIFF
--- a/Sources/FioriSwiftUICore/Utils/Formatter/CustomNumberFormatter.swift
+++ b/Sources/FioriSwiftUICore/Utils/Formatter/CustomNumberFormatter.swift
@@ -8,292 +8,290 @@
 //  without the express permission of SAP SE. The information contained herein may be changed
 //  without prior notice.
 //
-#if !os(visionOS)
-    import UIKit
+import UIKit
 
-    /**
-     This is an implementation of `FormattedStringEditing` on top of the regular iOS `NumberFormatter`.
-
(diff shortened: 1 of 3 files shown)


Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -389,7 +389,7 @@ protocol _IllustratedMessageComponent: _DetailImageComponent, _TitleComponent, _
 // sourcery: CompositeComponent
 protocol _InformationViewComponent: _IconComponent, _DescriptionComponent {}
 
-// sourcery: CompositeComponent, InternalComponent
+// sourcery: CompositeComponent
 protocol _TextInputInfoViewComponent: _InformationViewComponent, _CounterComponent {}
```
janhuachu, line 411: This _TextInputInfoViewComponent changed to public without header docs. Please add header docs so that the generated structs will have docs.

KevinZK: Added

PR #1311, opened 2025-10-28
feat: 🎸 add two joule colors and update chart1 color

dyongxu:


Sources/FioriThemeManager/Colors/ColorStyle.swift
```diff
@@ -1330,6 +1330,18 @@ public enum ColorStyle: String, CaseIterable {
...
+    /// (light variant) ![](https://dummyimage.com/48x18/470CED.png&text=+)          Hex value: 470CED
+    @available(watchOS, unavailable)
+    case jouleSelectedFill
+    
+    /// jouleSelectedLabel
+    /// (dark variant) ![](https://dummyimage.com/48x18/470CED.png&text=+)          Hex value: 470CED
+    /// (light variant) ![](https://dummyimage.com/48x18/E2D8FF.png&text=+)          Hex value: E2D8FF
+    @available(watchOS, unavailable)
+    case jouleSelectedLabel
+
```
KevinZK, line 1344: The allCases of ColorStyle need to add jouleSelectedLabel and jouleSelectedFill.
Otherwise, testLatestColorStyle() will go wrong.
In the case of Colors, these two colors will also be lost.

PR #1320, opened 2025-10-31
feat: 🎸 [IOSSDKBUG-1006]DateRangePicker: click optimization

janhuachu:
After a range was selected by user, user tapped another day outside of the range. The original range is cleared and the new date will be the starting date of a new range.

✅ Closes: IOSSDKBUG-1006

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/TabViewExample/TabViewDetailView.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/TabViewExample/TabViewDetailView.swift
@@ -70,6 +70,7 @@ struct TabViewDetailView: View {
                 .tag(index)
                 .tabItem {
                     Label(model.name, systemImage: model.imageName)
+                        .accessibilityAddTraits(self.selection == index ? [.isSelected] : [])
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


Sources/FioriSwiftUICore/_FioriStyles/DateRangePickerStyle.fiori.swift
```diff
@@ -28,11 +28,29 @@ class DateRangePickerModelObject: ObservableObject {
             self.selectedDates = newSelection
             self.tapCount = 1
         } else if self.selectedDates.count == 1, self.tapCount < 2 {
-            self.selectedDates = self.updateDateRange(self.selectedDates, newSelection)
+            if !newSelection.isEmpty {
+                self.selectedDates = self.updateDateRange(self.selectedDates, newSelection)
+            }
+            // If newSelection isEmpty means user tapped the same date.
+            // Since we supports range with 1 date, do not set selectedDates to empty.
```
dyongxu:
```suggestion
            // Since we support range with 1 date, do not set selectedDates to empty.
```

dyongxu (changes requested):
typo correction

PR #1333, opened 2025-11-07
fix: 🐛 [IOSSDKBUG-975] Edit illustration @available annotations

JuliaSchnSAP:


Sources/FioriThemeManager/FioriIcons/FioriIcon.swift
```diff
@@ -2842,677 +2842,677 @@ public enum FioriIcon {
         ///
         public static let userHasSignedUpSmall = Image(fioriName: "fiori.ill.user.has.signed.up.small")
 
-        // Removed illustrations
-        @available(swift, obsoleted: 5.0, message: "Illustration was removed")
+        // Illustrations that were removed in 25.11
+        @available(swift, obsoleted: 5.0, message: "Illustration was replaced", renamed: "addingColumnsMedium")
```
billzhou0223: Is there a specific reason to mark this api `obsoleted` from swift v5.0? Instead of just marking it as `unavailable`?

billzhou0223 (changes requested):

billzhou0223 (approved):

PR #1356, opened 2025-11-18
fix: 🐛 [IOSSDKBUG-1128] allow show/hide feedback footer in W.A

xiaoyu0722:


Sources/FioriSwiftUICore/AIWritingAssistant/InternalWAForm.swift
```diff
@@ -185,7 +186,7 @@ struct InternalWAForm: View {
     }
     
     @ViewBuilder func sectionFooter() -> some View {
-        if self.context.feedbackHandler != nil {
+        if !self.hideWAFeedbackFooter {
```
hengyi-zhang: I don't think the check `self.context.feedbackHandler != nil` should be removed

hengyi-zhang (changes requested):

Sources/FioriSwiftUICore/AIWritingAssistant/InternalWAForm.swift
xiaoyu0722: Using `hideWAFeedbackFooter` to show/hide this footer, even feedback handler is empty.
SDK will not maintain this logic anymore, if no handler, and footer is displayed, the tap action for vote also will be ignored.

Sources/FioriSwiftUICore/AIWritingAssistant/InternalWAForm.swift
```diff
@@ -212,3 +213,23 @@ struct InternalWAForm: View {
...
+        get { self[HideWAFeedbackFooterKey.self] }
+        set { self[HideWAFeedbackFooterKey.self] = newValue }
+    }
+}
+
+public extension View {
+    /// Hide feedback section footer for writing assistant.
+    /// - Parameter hide: A boolean value to indicate if hide/show the feedback footer.
+    /// - Returns: A new view with hidden or shown feedback footer for writing assistant.
+    func hideWAFeedbackFooter(_ hide: Bool = true) -> some View {
```
billzhou0223: Don't use abbreviation in a public api which is not clear enough.

billzhou0223 (changes requested):

dyongxu (approved):

PR #1358, opened 2025-11-18
fix: 🐛 [IOSSDKBUG-1171] [acc] add state acc label for SPI

xiaoyu0722:


hengyi-zhang (approved):

Sources/FioriSwiftUICore/Views/StepProgressIndicator/StepsStyles.swift
```diff
@@ -60,18 +60,51 @@ struct StepButtonStyle: ButtonStyle {
...
+                            line: self.generateLine(stepConfig),
+                            isTitleEmptyView: self.isTitleEmptyView,
+                            top: self.top,
+                            bottom: self.bottom,
+                            leading: self.leading,
+                            trailing: self.trailing,
+                            horizontalSpacing: self.horizontalSpacing,
+                            verticalSpacing: self.verticalSpacing,
+                            lineSize: self.lineSize)
+            self.stateAccessibilityLabel(by: self.stateAccessibilityLabelValue)
```
billzhou0223: Why adding this dummy view with accessibility label instead of setting the label on the `InnerSingleStep` directly?

xiaoyu0722: We wrap everything into `View`, can not access acc label of `InnerSingleStep` directly.
I did not find other way to add some extra acc label for a `View`.
Like a view with original acc label `Step A`, I want to update it to `Step A` + `active`. But currently I can't get `Step A` for a view.

xiaoyu0722: I remove the dummy view and attach the extra accessibility to line.

zzchao-1999 (approved):

PR #1362, opened 2025-11-19
fix: 🐛 [IOSSDKBUG-1025] Add acc support to icon in BannerMessage

hengyi-zhang:


hengyi-zhang:
> Make text localized string

The texts in framework already use localized string. Only the texts in example code don't.

dyongxu (approved):

PR #1365, opened 2025-11-20
fix: 🐛 [IOSSDKBUG-1072] ACC- SortFilter list picker searched information not readout

leoxiao-sap:


Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
```diff
@@ -957,6 +957,17 @@ struct ListPickerDestinationContent<Data: RandomAccessCollection, ID: Hashable,
...
+        
+        if UIAccessibility.isVoiceOverRunning {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                var formatString = "no matching data is available".localizedFioriString()
+                if !filteredResult.isEmpty {
+                    formatString = String(format: "%d number of items found".localizedFioriString(), filteredResult[0].1.count)
+                }
+                UIAccessibility.post(notification: .announcement, argument: formatString)
+            }
+        }
```
billzhou0223, line 992: It will be called every time view body is recomputed, which may happen more than once. I don't think this is the proper place to handle accessibility. Maybe we should handle it in modifiers like `onAppear()` which is only called once. Or we may use SwiftUI specific accessibility modifiers instead of UIAccessibility api.

I will let @xiaoyu0722 decide since he owns this component.

leoxiao-sap: @billzhou0223 Yes, reading out should be triggered for every character input or deletion during search, as the search results are constantly changing. cc: @xiaoyu0722

xiaoyu0722: I think we do not need to support this acc relevance between search field and items.
After search is active, only hovering on the item, then it should be announced.

PR #1370, opened 2025-11-24
fix: 🐛 [IOSSDKBUG-1206]add more area for focus indicator

KevinZK:
Screenshot 2025-11-24 at 13 33 02

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/BarcodeScanner/VisionKitMultipleBarcodesScannerExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/BarcodeScanner/VisionKitMultipleBarcodesScannerExample.swift
@@ -1,0 +1,102 @@
+import FioriSwiftUICore
+import Foundation
+import OSLog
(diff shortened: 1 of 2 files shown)


billzhou0223 (changes requested):
1. Did you get these hardcoded numbers from designer?
2. Should we change the original frame or apply a new frame outside of the original frame? This will create more white space around the indicator. Please confirm with designer if this is the correct design.

dyongxu (commented):
Please check if this size is the same as UIKit implementation.

KevinZK:
> 1. Did you get these hardcoded numbers from designer?
> 2. Should we change the original frame or apply a new frame outside of the original frame? This will create more white space around the indicator. Please confirm with designer if this is the correct design.


A fixed frame has been set for the indicator in both components. Setting the min/max of the frame will not result in more white space

dyongxu:
@billzhou0223 Do you have pending requested changes ?

dyongxu (approved):

PR #1378, opened 2025-11-26
fix: 🐛 [IOSSDKBUG-1073] ACC- SortFilter- ‘List Multiple’ item should contains sub-controls without causing duplicate readouts

leoxiao-sap:


Sources/FioriSwiftUICore/Views/SortFilter/SortFilterCFGItemContainer.swift
```diff
@@ -283,22 +283,34 @@ extension SortFilterCFGItemContainer: View {
             Text(self._items[r][c].picker.name)
                 .font(.fiori(forTextStyle: .subheadline, weight: .bold, isItalic: false, isCondensed: false))
                 .foregroundColor(Color.preferredColor(.primaryLabel))
+                .accessibilityElement()
```
billzhou0223: Is setting the Text an accessibility element explicitly really needed? Should the Text be accessible by default?

leoxiao-sap: Hi @billzhou0223 You are right, I have updated the code, thank you!

billzhou0223 (approved):

PR #1392, opened 2025-12-02
fix: 🐛 [IOSSDKBUG-1257] ACC-Attachment control state of the UI is not read out.

leoxiao-sap:
dev result
image
image

### DIFF
--- a/Sources/FioriSwiftUICore/_localization/en.lproj/FioriSwiftUICore.strings
+++ b/Sources/FioriSwiftUICore/_localization/en.lproj/FioriSwiftUICore.strings
@@ -196,11 +196,47 @@
 /* XBUT: timeline preview component see all label */
 "See All (%d)" = "See All (%d)";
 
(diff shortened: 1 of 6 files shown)


Sources/FioriSwiftUICore/_localization/en.lproj/FioriSwiftUICore.strings
```diff
@@ -394,6 +394,12 @@
 /* XACT: The accessibility hint for add button */
 "Double tap to add attachment" = "Double tap to add attachment";
 
+/* XACT: The accessibility hint for attachment section */
+"read only" = "read only";
+
+/* XACT: The accessibility hint for attachment section */
+"actions avaiable" = "actions avaiable";
```
dyongxu:
```suggestion
"actions available" = "actions available";
```

dyongxu (changes requested):
fix typo

PR #1397, opened 2025-12-03
fix: 🐛 [JIRA:0] Change CalendarDayView realization to avoid style override

hengyi-zhang:


Sources/FioriSwiftUICore/_FioriStyles/CalendarDayViewStyle.fiori.swift
```diff
@@ -97,79 +128,6 @@ extension CalendarDayViewFioriStyle {
...
-                return self.calendarItemTintAttributes[.title]?[.disabled] ?? .preferredColor(.quaternaryLabel)
-            case .outOfMonth:
-                return .preferredColor(.quaternaryLabel)
-            }
-        }
-        
-        /// Limit max type size to save more space to display all the contents
-        var scaleForSizeChange: Double {
-            sizeEnumToValue(dynamicTypeSize: self.dynamicTypeSize, limitMaxTypeSize: .accessibility1)
-        }
```
xiaoyu0722, line 171: I see you are still using this property?

hengyi-zhang: Yeah, this property is used to limit the max font size of the title or subtitle.

PR #1401, opened 2025-12-04
fix: 🐛 [IOSSDKBUG-1277] ACC-Text input info view - screen reader is reading as hint text 4/200 text field which is not meaningful

leoxiao-sap:
dev results:

iShot_2025-12-04_14 54 52
iShot_2025-12-04_14 54 09
iShot_2025-12-04_14 51 59

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/BarcodeScanner/VisionKitMultipleBarcodesScannerExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/BarcodeScanner/VisionKitMultipleBarcodesScannerExample.swift
@@ -1,0 +1,102 @@
+import FioriSwiftUICore
+import Foundation
+import OSLog
+import SwiftUI
+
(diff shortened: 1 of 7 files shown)


Sources/FioriSwiftUICore/_localization/en.lproj/FioriSwiftUICore.strings
```diff
@@ -310,6 +310,12 @@
...
 "Text field, Double tap to edit" = "Text field, Double tap to edit";
 
+/* XACT: The accessibility hint for Text Input Views, editable */
+"%d character left" = "%d character left.";
+
+/* XACT: The accessibility hint for Text Input Views, editable */
+"%d characters left" = "%d characters left.";
+
 /* XACT: The accessibility hint for Text Input Views, editing */
 "Text field, is editing" = "Text field, is editing";
```
dyongxu:
```suggestion
"Text field, is editing" = "Text field, in edit mode";
```

dyongxu (approved):

PR #1409, opened 2025-12-08
feat: 🎸 [3079] inline mode support for AI feedback

xiaoyu0722:


Apps/Examples/Examples/FioriSwiftUICore/AIUserFeedback/AIUserFeedbackExample.swift
```diff
@@ -58,13 +59,16 @@ struct AIUserFeedbackExample: View {
                 }
             }
             
+            self.showFeedback(mode: .inline)
```
dyongxu: In the func showFeedback(), update the title to include .inline, so that the user easily understand this is the inline AIUserFeedback example.

Sth like:
                         title: {
                                    if mode == .inline {
                                        Title(title: "How was your AI experience? (Inline Mode)")
                                    }
                                    else {
                                        Title(title: "How was your AI experience?")
                                    }
                              },

dyongxu (changes requested):
update test description

Apps/Examples/Examples/FioriSwiftUICore/AIUserFeedback/AIUserFeedbackExample.swift
xiaoyu0722: I did some test, seems AIUserFeedback still got some bugs, I will update this PR after it's fixed.

PR #1411, opened 2025-12-09
fix: 🐛 [JIRA:0] Update scrollPosition when tap status bar in CalendarView

hengyi-zhang:
Using .onScrollGeometryChange modifier to update scrollPosition when tap status bar to  scroll the ScrollView to top in CalendarView.

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/CalendarViewStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/CalendarViewStyle.fiori.swift
@@ -222,21 +222,10 @@ public struct CalendarViewBaseStyle: CalendarViewStyle {
                 configuration.model.selectedDates = [date]
             }
         } else {
-            let calendar = Calendar.autoupdatingCurrent
             if configuration.model.calendarStyle == .rangeSelection {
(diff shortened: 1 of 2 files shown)


Sources/FioriSwiftUICore/_FioriStyles/CalendarViewStyle.fiori.swift
```diff
@@ -182,6 +182,24 @@ public struct CalendarViewBaseStyle: CalendarViewStyle {
                     }
                 }
             }
+            .ifApply(true, content: {
```
xiaoyu0722: Why don't you move this to fioriScrolOffsetReeader?
```swift
func fioriScrollOffsetReader(_ offset: @escaping (CGPoint) -> Void) -> some View {
if #available(iOS 18.0, *) {
        return onScrollGeometryChange(for: CGPoint.self) { geometry in
            geometry.contentOffset
        } action: { _, newValue in
            offset(newValue)
        }
        } else {
          return self
        }
    }
```

hengyi-zhang: Fixed!

PR #1422, opened 2025-12-15
fix: 🐛 Improved card headerAction accessibility

zzchao-1999:


Sources/FioriSwiftUICore/_FioriStyles/CardStyle.fiori.swift
```diff
@@ -225,6 +225,7 @@ extension CardFioriStyle {
         
         func makeBody(_ configuration: HeaderActionConfiguration) -> some View {
             HeaderAction(configuration)
+                .accessibilityHint(NSLocalizedString("Header action", tableName: "FioriSwiftUICore", bundle: Bundle.accessor, comment: ""))
```
dyongxu:
```suggestion
                .accessibilityHint(NSLocalizedString("Card Header action", tableName: "FioriSwiftUICore", bundle: Bundle.accessor, comment: ""))
```

dyongxu: Will it make it clearer to add "Card" here to make it more explicit ? Similar to strings like "Card Subtitle".

dyongxu (changes requested):
may want to update localized string

PR #1431, opened 2025-12-17
fix: 🐛 [IOSSDKBUG-1587] Workaround for scrolling problem related to FioriButton in iOS 26.0

hengyi-zhang:


Sources/FioriSwiftUICore/FioriButton/FioriButton.swift
```diff
@@ -150,19 +150,36 @@ public struct FioriButton: View {
...
-                    Color.clear.contentShape(Rectangle()).simultaneousGesture(self.createGesture(proxy.size))
-                })
-                .setOnChange(of: self.isSelectionPersistent) {
-                    self._state = .normal
-                }
-                .ifApply(self.isLoading) { view in
-                    view.opacity(0.25)
+        // For menu use case, fioriButton should be based on Button
+        if #available(iOS 26.0, *) {
+            VStack {
```
billzhou0223, line 155: Why do we need this extra VStack?

dyongxu (approved):

Sources/FioriSwiftUICore/FioriButton/FioriButton.swift
hengyi-zhang: When tapping the Button quickly in List, the ButtonStyleConfiguration’s isPressed property don't change properly, always false. Wrapped in VStack is a workaround.

PR #1436, opened 2025-12-22
fix: 🐛[IOSSDKBUG-1313] [acc] Set the Illustrated Message description to allow multiple lines

zuopengchi:
In large text size, to display the full description without it being cut off, increase the height from 300 to 800.
Simulator Screenshot - iPad (A16) - 2025-12-22 at 14 40 18


Apps/Examples/Examples/FioriSwiftUICore/IllustratedMessage/IllustratedMessageExample.swift
```diff
@@ -113,7 +113,7 @@ struct IllustratedMessageExample: View {
                         }, title: {
                             Text("Illustrated Message Title with long text for two lines")
                         }, description: {
-                            hasDescription ? Text("This is description text for the Illustrated Message component") : nil
+                            hasDescription ? Text("This is description text for the Illustrated Message component").lineLimit(nil) : nil
```
billzhou0223, line 116: If unlimited number of lines is required by the spec as default configuration, we should set it in the SDK.

PR #1443, opened 2025-12-24
fix: 🐛 [IOSSDKBUG-1282]: Press up arrow key within the ‘How was your Ai experience’ nothing happens

leoxiao-sap:


dyongxu (commented):
The "up" and "down" arrow is not supposed to show more contents. 
The default behavior for them is supposed to move to the next item.
The "space" key is supposed to activate the action.

billzhou0223 (approved):

dyongxu (changes requested):
I don't think we need this fix.

leoxiao-sap:
> The "up" and "down" arrow is not supposed to show more contents. The default behavior for them is supposed to move to the next item. The "space" key is supposed to activate the action

Hi @dyongxu, Only the "down" arrow is supposed to move to the next item, "up" arrow is supposed to show success toast and dismiss this panel. Attached the before video:
Before:
https://github.com/user-attachments/assets/1d45a01a-ad27-4674-a95e-1fbc63a65f36
What we supposed is like following:
https://github.com/user-attachments/assets/85f4fd58-e58c-45c6-b527-e6c25590b4c4

dyongxu:
@leoxiao-sap  Based on our discussion, pls close this pull request.

PR #1470, opened 2026-01-06
feat: 🎸 selected value at top for ListPicker singleSelection

dyongxu:
Request from HCPSDKFIORIUIKIT-3113: 

https://github.com/user-attachments/assets/3ce437c8-d90c-4fc5-9096-2a08b98dfbcd

 For Single-select list picker , make selected option show up at the top.

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
@@ -889,25 +889,19 @@ struct ListPickerDestinationContent<Data: RandomAccessCollection, ID: Hashable, 
     }
         
     @ViewBuilder func generateSection(by serialData: [Data.Element]) -> some View {
-        // Only sort for single selection mode to show selected item first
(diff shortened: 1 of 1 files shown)


Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
```diff
@@ -889,7 +889,25 @@ struct ListPickerDestinationContent<Data: RandomAccessCollection, ID: Hashable,
...
+            // Selected items come first
+            if isSelected1, !isSelected2 {
+                return true
+            } else if !isSelected1, isSelected2 {
+                return false
+            } else {
+                // Maintain original order for items with same selection state
+                return false
+            }
+        } : serialData
```
billzhou0223: Since there is only one selected item. The following logic should be more efficient and readable. Sort function takes O(nlogn) time but the following takes O(n)

```
let reorderedData = serialData
let selectedItemIndex = reorderedData.firstIndex { isItemSelected($0.id) }
let item = reorderedData.remove(at: selectedItemIndex)
reorderedData.insert(item, at: 0)
```

dyongxu: updated

PR #1471, opened 2026-01-07
fix: 🐛 [IOSSDKBUG-1337] wrap long text of action button in cardfooter

zzchao-1999:


Sources/FioriSwiftUICore/_FioriStyles/CardFooterStyle.fiori.swift
```diff
@@ -90,8 +90,12 @@ private struct CardFooterLayout: Layout {
         
         cache.clear()
         
-        let subViewSizes = subviews.reversed().map {
-            $0.sizeThatFits(proposal)
+        // Use width-constrained proposal when calculating subview sizes to enable text wrapping
```
shengxu7, line 93: @zzchao-1999  This is not necessary since maxButtonWidth is always nil so far.

Sources/FioriSwiftUICore/_FioriStyles/CardFooterStyle.fiori.swift
```diff
@@ -149,6 +153,7 @@ private struct CardFooterLayout: Layout {
             // check if overflow should be shown
             if numToShow < numButtons {
                 requiredFinalWidth += min(self.maxButtonWidth, overflowSize.width) + theSpacing
+                maxHeight = max(maxHeight, overflowSize.height)
```
shengxu7, line 156: @zzchao-1999 This is not needed since line 198 does this. 
if numToHide > 0 {
            maxHeight = max(maxHeight, overflowSize.height)
        }

shengxu7 (changes requested):
@zzchao-1999 What problem do you try to solve? Can you provide a before and after screenshots. Original version should support text wrapping for buttons.

dyongxu:
Simulator Screenshot - iPad Pro 13-inch (M5) - 2026-01-08 at 16 06 02

zzchao-1999:
The issue has been fixed by below checkin:
https://github.com/SAP/cloud-sdk-ios-fiori/pull/1465.

PR #1473, opened 2026-01-07
fix: 🐛 [IOSSDKBUG-1178] [acc] Add No Line Limit option in KPI Header example

zuopengchi:
Add an "KPI Caption Unlimited Lines" toggle to the KPI Header example. The default value of it is false. When enabled, it removes the line limit for the KPI item caption.
Screenshot 2026-01-07 at 16 33 22
Screenshot 2026-01-07 at 16 33 35


dyongxu:
@zuopengchi It is a different issue than IOSSDKBUG-1178:  the "Downloading" is stretched outside of the circle. We may need to fix it with another PR.

PR #1482, opened 2026-01-12
feat: 🎸 [HCPSDKFIORIUIKIT-3112]List Picker support Readonly

KevinZK:
For List Picker Form Cell: support Read-only state

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/BannerMessage/BannerMultiMessageCustomInitExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/BannerMessage/BannerMultiMessageCustomInitExample.swift
@@ -173,6 +173,13 @@ struct BannerMultiMessageCustomInitExample: View {
                             }
                         }, bannerMultiMessages: self.$bannerMultiMessages)
                             .presentationDetents([.medium, .large])
+                            .presentationBackground(.clear)
+                            .overlay(
+                                ContainerRelativeShape()
+                                    .inset(by: 2)
+                                    .stroke(Color.red, lineWidth: 4)
+                                    .ignoresSafeArea(.container, edges: .all)
+                            )
                     }
                 }
                 .listRowSeparator(.hidden)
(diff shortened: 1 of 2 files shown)


Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
```diff
@@ -47,13 +47,16 @@ struct ListPickerItemExample: View {
     @State var showsErrorMessage = false
     @State var showsPrompt = false
     @State var showAINotice: Bool = false
-    
+    var value: AttributedString? {
+        self.state == .readOnly ? "Value" : nil
```
dyongxu: update "Value" to get the actual selected values from the listPickerItem, so it will reflect clearly how it looks. (Especially, to check the layout when it is multiple selection allowed, and it is vertical)

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -1063,6 +1063,8 @@ protocol _AvatarStackComponent: _AvatarsComponent, _AvatarsTitleComponent {}
 ///         Text("title")
 ///     }, value: {
 ///         Text("value")
+///     }, description: {
```
dyongxu, line 1089: Update the API Doc, how the listPicker will be read-only using controlState.
Currently, it is not clear to App developer how to use it.

dyongxu (changes requested):
Sample Code update in TestApp.
Update API doc.

Apps/Examples/Examples/FioriSwiftUICore/FormCells/ListPickerItemExample.swift
KevinZK: In the read-only mode, it is impossible to select values.

dyongxu: The value is not selectable. But the  listpicker should be able to configure its values before hand in App code, and display the values as "read only"

KevinZK: fixed

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
KevinZK: updated

PR #1484, opened 2026-01-12
fix: 🐛 [IOSSDKBUG-1606] hide time for date picker in sortfilter

xiaoyu0722:
|date & time|date only|
|-|-|
|Simulator Screenshot - iPad Pro 13-inch (M5) - 2026-01-12 at 16 50 08|Simulator Screenshot - iPad Pro 13-inch (M5) - 2026-01-12 at 16 50 10|

### DIFF
--- a/Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
+++ b/Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
@@ -1168,7 +1168,7 @@ public extension SortFilterItem {
         ///   - value: Date value of the DateTimeItem.
         ///   - formatter: Formatter string for the date value.
         ///   - icon: Icon name for the DateTimeItem. Default value is `nil`.
-        ///   - components: Date picker components to display in view hierarchy. Default value is `[.hourAndMinute, .date]`.
+        ///   - components: Date picker components to display in view hierarchy. If set to `.date`, the "Hour and Minute" field will not show up in the picker. Default value is `[.hourAndMinute, .date]`.
         public init(id: String = UUID().uuidString,
                     name: String,
                     value: Date?,


Apps/Examples/Examples/FioriSwiftUICore/SortFilter/SortFilterExample.swift
```diff
@@ -33,7 +33,8 @@ struct SortFilterExample: View {
                 }
                 return (.fiori, "")
             })), showsOnFilterFeedbackBar: true),
-            .datetime(item: .init(name: "Start Date", value: Date(), formatter: "yyyy-MM-dd HH:mm", icon: "calendar"), showsOnFilterFeedbackBar: true)
+            .datetime(item: .init(name: "Start Date", value: Date(), formatter: "yyyy-MM-dd HH:mm", icon: "calendar"), showsOnFilterFeedbackBar: true),
+            .datetime(item: .init(name: "Start Date", value: Date(), formatter: "yyyy-MM-dd", icon: "calendar", components: .date), showsOnFilterFeedbackBar: true)
```
dyongxu, line 37: .datetime(item: .init(name: "Start Date", value: Date(), formatter: **_"yyyy-MM-dd HH:mm",_** icon: "calendar", components: .date),

Will the above show the time or not?
Is it doable not introducing the new property "components" ?  SDK can inspect the formatter, to decide to show the time or not

xiaoyu0722: Hi @dyongxu 
SDK can check if there is time part in formatter, but I'm not sure if it's a good idea to bind it to picker style.

Like if there is only hour part without minute, I don't think, current implementation will support this. Or there is second part, also is not supported. Current date picker only support `date` and `hourAndMinute`.

But if you insist, I can re-code this part.

Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
```diff
@@ -1136,15 +1136,65 @@ public extension SortFilterItem {
...
+        }
+        
+        /// Convenience init to create DateTimeItem with custom date picker components.
+        /// - Parameters:
+        ///   - id: Unique identifier for the DateTimeItem.
+        ///   - name: Name of the DateTimeItem.
+        ///   - value: Date value of the DateTimeItem.
+        ///   - formatter: Formatter string for the date value.
+        ///   - icon: Icon name for the DateTimeItem. Default value is `nil`.
+        ///   - components: Date picker components to display in view hierarchy. Default value is `[.hourAndMinute, .date]`.
```
dyongxu: If set to .Date, the "Hour and Minute" field will not show up in the Picker.

dyongxu: Update doc to make it clear how to use it.

Apps/Examples/Examples/FioriSwiftUICore/SortFilter/SortFilterExample.swift
dyongxu: as per discussion, a new implementation will be done.

PR #1508, opened 2026-01-21
fix: 🐛 [IOSSDKBUG-1627] AI Writing Assistant Tapped State Update

leoxiao-sap:
1. Fixed the background color not right issue on iOS 26.1
2. Fixed the issue where long-pressing and releasing did not trigger the selection event.


Sources/FioriSwiftUICore/AIWritingAssistant/InternalWAForm.swift
```diff
@@ -106,11 +106,14 @@ struct InternalWAForm: View {
     
     @ViewBuilder var listContent: some View {
         let sections = self.menus.map { WritingToolSection(menus: $0) }
-        List {
+        List(selection: self.$context.selection) {
```
xiaoyu0722, line 109: If use this again, does setting selection manually needed in tap gesture?

leoxiao-sap: @xiaoyu0722 Yes, otherwise the tap gesture won't change selection because it interferes with the list's default selection behavior.  When the tap gesture is added, the default selection behavior will only be triggered when the row is long-pressed.

PR #1517, opened 2026-01-26
fix: 🐛 [IOSSDKBUG-1139] [acc] Add a flag enabling the detail image default style in ContactItem

zuopengchi:
If user don’t want the default detail image style (e.g., when passing a non-image view), set the environment variable removeDetailImageDefaultStyle to true.

Screenshot 2026-01-26 at 14 54 14
Screenshot 2026-01-26 at 14 54 24

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/ContactItemStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/ContactItemStyle.fiori.swift
@@ -165,12 +165,12 @@ extension ContactItemFioriStyle {
     }
 
     struct DetailImageFioriStyle: DetailImageStyle {
-		@Environment(\.removeDetailImageDefaultStyle) private var removeDetailImageDefaultStyle
+        @Environment(\.removeDetailImageDefaultStyle) private var removeDetailImageDefaultStyle
         let contactItemConfiguration: ContactItemConfiguration
 
         func makeBody(_ configuration: DetailImageConfiguration) -> some View {
             DetailImage(configuration)
-                .ifApply(!removeDetailImageDefaultStyle) {
+                .ifApply(!self.removeDetailImageDefaultStyle) {
                     $0.frame(width: 45, height: 45, alignment: .center)
                         .clipShape(Circle())
                 }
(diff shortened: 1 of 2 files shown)


Sources/FioriSwiftUICore/_FioriStyles/ContactItemStyle.fiori.swift
```diff
@@ -169,8 +169,10 @@ extension ContactItemFioriStyle {
 
         func makeBody(_ configuration: DetailImageConfiguration) -> some View {
             DetailImage(configuration)
-                .frame(width: 45, height: 45, alignment: .center)
-                .clipShape(Circle())
+                .ifApply(self.contactItemConfiguration.usesDetailImageDefaultStyle) {
+                    $0.frame(width: 45, height: 45, alignment: .center)
+                        .clipShape(Circle())
+                }
```
xiaoyu0722, line 176: Hi @dyongxu @zuopengchi  What about remove these directly?
And first, frame should be set by developer but not sdk, like `ObjectItem`.
Second, use a environment value to check the clip shape, something like
```swift
.clipShape(isDetailImageCircular ? AnyShape(Circle()) : AnyShape(Rectangle()))
```

Sources/FioriSwiftUICore/_ComponentProtocols/CompositeComponentProtocols.swift
```diff
@@ -1407,7 +1407,11 @@ protocol _ActivityItemComponent: _IconComponent, _SubtitleComponent {
 /// }
 /// ```
 // sourcery: CompositeComponent
-protocol _ContactItemComponent: _TitleComponent, _SubtitleComponent, _DescriptionComponent, _DetailImageComponent, _ActivityItemsComponent {}
+protocol _ContactItemComponent: _TitleComponent, _SubtitleComponent, _DescriptionComponent, _DetailImageComponent, _ActivityItemsComponent {
+    /// This property determines whether the DdetailImage is displayed using the default style. The default is true.
```
xiaoyu0722:
```suggestion
    /// This property determines whether the `detailImage` is displayed using the default style. The default is true.
```

xiaoyu0722 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/ContactItemStyle.fiori.swift
dyongxu: @xiaoyu0722 @zuopengchi That is my preference, if we can do this at App level.  I discussed with @zuopengchi  yesterday, but he said not doable. So I asked him to discuss with the team to see any other options.

zuopengchi: @dyongxu 
After disccussing it with @xiaoyu0722 , I see two options. 
One, if the user doesn’t care about changing the frame of the existing detail image in ContactItem, you can remove all of the image’s default styling and rely on app-level styling.
Two, use my current fix to handle the existing detail image, though it may not align with SwiftUI conventions.
Which one do you prefer?

dyongxu: @xiaoyu0722 I think it is ok for SDK to set the frame, however they should be just the default. There should be a way for Application to totally ignore our default frame. They may even replace the DetailImageView as a Card view, or Text View, etc.
@zuopengchi  My vote is option 1. SDK should have a default image style. So we should not remove it totally in SDK.

zuopengchi: After discussing with @xiaoyu0722 , we decided to use an environment variable instead of adding a new property

xiaoyu0722 (approved):

Sources/FioriSwiftUICore/_FioriStyles/ContactItemStyle.fiori.swift
```diff
@@ -185,3 +188,24 @@ extension ContactItemFioriStyle {
...
+		get { self[RemoveDetailImageDefaultStyleKey.self] }
+		set { self[RemoveDetailImageDefaultStyleKey.self] = newValue }
+	}
+}
+
+public extension View {
+	
+	/// Suppot for flexible style of`DetailImage`.
+	/// - Parameter value: A boolean value indicating if the default style is removed. Default value is `false`.
+	/// - Returns: The default style of `DetailImage`is removed or not.
```
xiaoyu0722: Typo issue and space needed between words.

PR #1520, opened 2026-01-28
fix: 🐛 [1661] render issue for dynamic acc setting in W.A

xiaoyu0722:
1. sync toolbar item color setting (Dynamic color seems failed to get correct trait collection sometimes)
2. For `HexColor`, `elevatedContrastLight` is not configured for some colors. 
But like popover, sheet or alert some elevated view will not get right contrast color style, so I update it by contrast color if there is one.

### DIFF
--- a/.hyperspace/pull_request_bot.json
+++ b/.hyperspace/pull_request_bot.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "https://devops-insights-pr-bot.cfapps.eu10-004.hana.ondemand.com/schema/pull_request_bot.json",
+  "features": {
+    "control_panel": false,
+    "summarize": {
+      "auto_generate_summary": true,
+      "auto_insert_summary": true,
+      "use_custom_summarize_prompt": false,
+      "use_custom_summarize_output_template": false
+    },
+    "review": {
+      "auto_generate_review": true
+    }
+  }
+}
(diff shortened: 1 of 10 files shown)


Sources/FioriSwiftUICore/AIWritingAssistant/WATextInput.swift
```diff
@@ -263,31 +264,36 @@ struct WATextInputModifier: ViewModifier {
...
+        self.formView
+            .frame(idealWidth: 400, idealHeight: 400)
+            .environmentObject(self.context)
+            .preferredColorScheme(self.colorScheme)
+            .presentationCompactAdaptation(.sheet)
+            .presentationDetents([.fraction(0.5)])
+            .interactiveDismissDisabled()
+            .presentationDragIndicator(.hidden)
+            .disabled(self.context.inProgress)
+            .toastMessage(isPresented: self.$context.showFeedbackSuccessToast, title: "Thank you for your feedback", duration: 3)
```
dyongxu:
```suggestion
            .toastMessage(isPresented: self.$context.showFeedbackSuccessToast, title: "Thank you for your feedback".localizedFioriString(), duration: 3)
```
1. update other places where "Thank you for your feedback" is used , to use localized string.
2. There could be other strings in WA, which do not use localized string yet .
3. Check https://github.com/SAP/cloud-sdk-ios-fiori/pull/1521, if it makes to move the color fix code to be in the Theming Manager side.

PR #1521, opened 2026-01-28
fix: 🐛 [IOSSDKBUG-1654] Adjust color variants default value in HexColor

hengyi-zhang:
Adjust color variants (.elevatedContrastLight & .elevatedContrastDark) default value in HexColor

### DIFF
--- a/.hyperspace/pull_request_bot.json
+++ b/.hyperspace/pull_request_bot.json
@@ -1,0 +1,15 @@
+{
+  "$schema": "https://devops-insights-pr-bot.cfapps.eu10-004.hana.ondemand.com/schema/pull_request_bot.json",
+  "features": {
+    "control_panel": false,
+    "summarize": {
+      "auto_generate_summary": true,
+      "auto_insert_summary": true,
+      "use_custom_summarize_prompt": false,
+      "use_custom_summarize_output_template": false
+    },
+    "review": {
+      "auto_generate_review": true
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


Sources/FioriThemeManager/Colors/HexColor.swift
```diff
@@ -38,22 +38,24 @@ public struct HexColor: Hashable {
...
+        if let color = contrastDarkColor {
+            self.colors.updateValue(color, forKey: .contrastDark)
             self.colors.updateValue(color, forKey: .elevatedContrastDark)
         }
         
-        if let color = contrastLightColor {
-            self.colors.updateValue(color, forKey: .contrastLight)
+        if let color = elevatedLightColor {
+            self.colors.updateValue(color, forKey: .elevatedLight)
+            self.colors.updateValue(color, forKey: .elevatedContrastLight)
```
xiaoyu0722, line 53: Hi @dyongxu  When `elevatedContrastLightColor` is missed in color palette.
Which one should be used for it? `elevatedLightColor` or `contrastLightColor`

dyongxu: If elevatedContrastLightColor is missing,  elevatedLightColor should be used ( if provided )
If elevatedLightColor is missing as well, then use contrastLightColor (if provided)

Elevated Color variant can be observed when the view is presented as a modal.  iOS system will trigger to apply the Elevated Color variant.

hengyi-zhang: The new code conforms the logic.

PR #1539, opened 2026-02-03
fix: 🐛 [IOSSDKBUG-1665] Sidebar can't work well on iOS 26.1 and 26.2

JunSong-SH:
fix: 🐛 [[IOSSDKBUG-1665](https://jira.tools.sap/browse/IOSSDKBUG-1665)] Sidebar can't work well on iOS 26.1 and 26.2

Found the issue during investigate the A11Y issue. The Sidebar doesn't work on iOS 26.1 and 26.2. When a user clicks a sidebar item, the detail page doesn't update and the sidebar item isn't highlighted. It works fine on iOS 26.0.

After investigating, I found the root cause: the .navigationDestination modifier no longer works with NavigationLink inside NavigationSplitView on iOS 26.1 and 26.2. As a result, the Sidebar no longer uses NavigationLink and .navigationDestination; instead, it uses plain Button controls internally. Consumers of the Sidebar can use the exposed selection state to determine the selected item and display the corresponding detail view in the detail pane of the NavigationSplitView.

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/ProfileHeader/ProfileHeaderAnimatableExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/ProfileHeader/ProfileHeaderAnimatableExample.swift
@@ -15,13 +15,16 @@ struct ProfileHeaderAnimatableExample: View {
             HStack(spacing: 30) {
                 Button(action: {}, label: {
                     Image(systemName: "mail")
+                        .frame(minWidth: 44, minHeight: 44)
                 })
                 Button(action: {}, label: {
                     Image(systemName: "message")
+                        .frame(minWidth: 44, minHeight: 44)
                 })
                 
                 Button(action: {}, label: {
                     Image(systemName: "phone")
+                        .frame(minWidth: 44, minHeight: 44)
                 })
             }
         }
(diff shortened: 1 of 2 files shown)


Sources/FioriSwiftUICore/_FioriStyles/SideBarStyle.fiori.swift
```diff
@@ -145,11 +132,14 @@ public struct SideBarBaseStyle: SideBarStyle {
...
+                            withAnimation {
+                                configuration.selection = item
+                            }
+                        } label: {
                             configuration.item(bindableItem).typeErased
-                        }.accessibilityAction(named: "Select") {}
+                        }
+                        .buttonStyle(.plain)
                     } else {
                         // in case the navigationDestination modifier of the NavigationLink can't work, to set the selection item when the NavigationLink was clicked or enter key pressed or double-tapping with VoiceOver
```
dyongxu: update comments here since NavigationLink is not used anymore after the fix

Sources/FioriSwiftUICore/_FioriStyles/SideBarStyle.fiori.swift
```diff
@@ -145,11 +132,14 @@ public struct SideBarBaseStyle: SideBarStyle {
...
-                        NavigationLink(value: item) {
+                        Button {
+                            withAnimation {
+                                configuration.selection = item
+                            }
+                        } label: {
                             configuration.item(bindableItem).typeErased
-                        }.accessibilityAction(named: "Select") {}
+                        }
+                        .buttonStyle(.plain)
```
dyongxu, line 142: add accessibility support for the button

Sources/FioriSwiftUICore/_FioriStyles/SideBarStyle.fiori.swift
JunSong-SH: Fxied

Sources/FioriSwiftUICore/_FioriStyles/SideBarStyle.fiori.swift
JunSong-SH: Fixed

PR #1548, opened 2026-02-05
fix: 🐛 [IOSSDKBUG-1699] [iOS][FilterFeedbackBar] Issues on expand/collapse filter sheet

leoxiao-sap:
1. Break layout cycle caused by unstable presentation detents
2. Fix sheet open/dismiss glitch on iOS 18 and below

Dev results:
on iOS 18.5
https://github.com/user-attachments/assets/9be7780a-a803-46e8-8f0e-53775d3b28c0

On iPad
Simulator Screenshot - iPad (A16) - 2026-02-03 at 17 49 02
Simulator Screenshot - iPad (A16) - 2026-02-03 at 17 48 47

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/SideBar/EditableSideBarExample.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/SideBar/EditableSideBarExample.swift
@@ -42,9 +42,6 @@ struct EditableSideBarExample: View {
     var isPartialCustom: Bool = false
     let isPad: Bool = UIDevice.current.userInterfaceIdiom == .pad
 
-    var deviceModelObject = DeviceExampleModelObject()
-    let destinationView = DeviceDetail()
-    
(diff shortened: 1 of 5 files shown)


Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
```diff
@@ -118,6 +119,7 @@ struct CancellableResettableDialogNavigationForm<Title: View, CancelAction: View
             GeometryReader { proxy in
                 self.navigationBar()
                     .task {
+                        self.maxAvailableHeight = proxy.size.height
```
xiaoyu0722: This is same as navigationBarHeight, you can use it directly

leoxiao-sap: Updated, thanks

PR #1559, opened 2026-02-10
chore: 🤖 [3155|3154]removed deprecated code

KevinZK:
SwiftUI: remove deprecated code for FilterFeedbackBarButtonModel, FilterFeedbackBarItemModel and SortViewModel


Sources/FioriSwiftUICore/Views/OptionListPickerItem+View.swift
```diff
@@ -118,6 +119,118 @@ extension OptionListPickerItem: View {
      }
  }
  */
+/// Option list picker configuration for styling
+public struct OptionListPickerButtonConfiguration {
```
xiaoyu0722: Seems, these are used for deprecated components, I think we should remove them also.

xiaoyu0722 (changes requested):

Sources/FioriSwiftUICore/Views/OptionListPickerItem+View.swift
KevinZK: OptionListPickerButtonConfiguration used in new component, removed OptionListPickerItem

KevinZK:
Resolved conflicts

PR #1579, opened 2026-03-02
fix: 🐛 [IOSSDKBUG-1711] Support selecting today as default value in date time picker when opened

hengyi-zhang:
Add a environment 'dateTimePickerAutoSelected' to support selecting today as the default value in date time picker when opened

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/NoteFormViewStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/NoteFormViewStyle.fiori.swift
@@ -8,16 +8,16 @@ public struct NoteFormViewBaseStyle: NoteFormViewStyle {
     @Environment(\.isLoading) var isLoading
     @Environment(\.isAILoading) var isAILoading
     public func makeBody(_ configuration: NoteFormViewConfiguration) -> some View {
-        VStack(alignment: .leading) {
-            SkeletonLoadingContainer {
+        SkeletonLoadingContainer {
+            VStack(alignment: .leading) {
                 self.getPlaceholderTextEditor(configuration)
                     .focused(self.$isFocused)
                     .disabled(self.getDisabled(configuration))
             }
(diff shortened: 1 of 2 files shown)


hengyi-zhang:
> Before the user actually tapped a date(no matter it is today, or other days), the date picker's value field should be updated. e.g. , when it opens as "No date selected", it should remain as "No date selected" , unless the user taps. Otherwise, the user can not dismiss the picker without changing the value.

Added a parameter 'autoSelected' to separate from former behavior.

Sources/FioriSwiftUICore/_generated/StyleableComponents/DateTimePicker/DateTimePicker.generated.swift
```diff
@@ -51,6 +51,8 @@ public struct DateTimePicker {
     @Binding var pickerVisible: Bool
     /// This property indicates whether the separator is to be displayed. Default is false.
     let hidesSeparator: Bool
+    /// If selectedDate's initial value is nil, this property will make the selectedDate to be .now by default when the date picker opened.
```
dyongxu:
```suggestion
    /// If this property is true and selectedDate's initial value is nil,  selectedDate will be set to  .now by default when the date picker opens.
    /// Default value is false.
```

hengyi-zhang: Fixed!

dyongxu (approved):

dyongxu (changes requested):
try to avoid changing the init(), so that it can be backported into release branch

hengyi-zhang:
> Add a environment 'dateTimePickerAutoSelected'

Changed to use the environment 'dateTimePickerAutoSelected'.

hengyi-zhang:
> try to avoid changing the init(), so that it can be backported into release branch

Changed to use the environment 'dateTimePickerAutoSelected'.

PR #1583, opened 2026-03-05
chore: 🤖 [IOSSDKBUG-1530]Accessibility :the announcement of the NoteFormView case is not correct

KevinZK:
NoteformView has already set the content of accessibilityHint, 

The accessibility Label depends on the developer to set it manually. If it is not set, the Text field will be read out.

The component cannot internally obtain the placeholder string, making concatenation impossible. The component is designed with an accessibilityElement of "combine", so the accessibility support for the placeholder can only be implemented in the test case.


dyongxu (changes requested):
I am not sure if this is the right place for the fix. We may want to put the fix into SDK .
Ask Xiao Peng to have another review.

PR #1616, opened 2026-03-23
fix: 🐛 [IOSSDKBUG-1624] Set fixedSize to description in InformationView

hengyi-zhang:
Before:
Simulator Screenshot - iPhone 17 Pro - 2026-03-23 at 16 56 36

After:
Simulator Screenshot - iPhone 17 Pro - 2026-03-23 at 16 55 09

Appearance in dynamic type:
X Small:
Simulator Screenshot - iPhone 17 Pro - 2026-03-24 at 07 41 19

Accessibility 5:
Simulator Screenshot - iPhone 17 Pro - 2026-03-24 at 07 41 08

### DIFF
--- a/Apps/Examples/Examples/Info.plist
+++ b/Apps/Examples/Examples/Info.plist
@@ -70,7 +70,7 @@
 	<string>LaunchScreen</string>
 	<key>UIRequiredDeviceCapabilities</key>
 	<array>
-		<string>armv7</string>
+		<string>arm64</string>
 	</array>
 	<key>UISupportedInterfaceOrientations</key>
 	<array>


Sources/FioriSwiftUICore/_FioriStyles/InformationViewStyle.fiori.swift
```diff
@@ -9,6 +9,7 @@ public struct InformationViewBaseStyle: InformationViewStyle {
             HStack(alignment: .center, spacing: 8) {
                 configuration.icon
                 configuration.description
+                    .fixedSize(horizontal: false, vertical: true)
```
billzhou0223: Will fixed size break the dynamic type?

Please move style implementation into corresponding fiori style object, e.g. `DescriptionFioriStyle`.

billzhou0223 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/InformationViewStyle.fiori.swift
hengyi-zhang: The fixed size doesn't break the dynamic type.
Already moved it to DescriptionFioriStyle.

PR #1634, opened 2026-04-07
fix: 🐛 [0]iOS 26.4: DateTime popover clipped in FilterFeedback

KevinZK:
Bug Description:
  The DateTime item in FilterFeedbackBar displays with clipped/incorrectly sized popover content on iPad (iOS 26.4 simulator). The layout only corrects itself after the user interacts with the content. iPhone is unaffected.                                   
                                                                                                                                  
  Root Cause:                                                                                                                     
  On iPad, popover sizing relies on `.frame(idealHeight:)`. The idealHeight is calculated as `sheetContentHeight + bottomHeight + navigationBarHeight`, but several timing issues cause the first-frame value to be incorrect:                                     
  
  1. Delayed content height measurement: `sheetContentHeight` starts at 0 and is measured asynchronously via `GeometryReader` + `.task`. On the first frame, `idealHeight` is only ~50pt (just the navigation bar height), far too small for the `DatePicker` (~400pt), causing the popover to render clipped.                                                                                          
  2. Delayed navigation bar height measurement: The old code measured navigation bar height by creating a hidden NavigationStack and reading its geometry asynchronously via .task, which also wasn't available on the first frame. The hardcoded initial value  of 50 also didn't account for iOS 26 Liquid Glass navigation bar height changes.
  3. Incorrect bottom spacing: `horizontalSizeClass` was initialized to `.compact `unconditionally, causing the Apply button's bottom spacing on iPad to use the iPhone value (6pt instead of 16pt). Although `WindowTraitReader` corrects this later, the popover size is already determined on the first frame and does not readjust.
                                                                                                                                  
  Fix:                                                                                                                            
                                                                                              
  - When content has not yet been measured (`sheetContentHeight == 0`), pass `nil` for `idealHeight` instead of an incorrect small value, allowing SwiftUI to use intrinsic content sizing for the popover and avoiding first-frame clipping 
                      
  - Replace async navigation bar height measurement with static values based on device type and Liquid Glass state, eliminating the timing issue; remove the now-unnecessary hidden NavigationStack measurement method      
                                      
  - Upgrade content height measurement from `GeometryReader` + `.task `(one-shot) to `onGeometryChange `(iOS 18+, continuous tracking), supporting dynamic content height changes (e.g., DatePicker switching between 5-week and 6-week months) and handling iOS 26's  inflated first-pass height 
                                                                                                       
  - Initialize horizontalSizeClass based on device type to ensure correct spacing on the first frame for iPad
                                                                                                                                                                                                                         
  - Remove `.frame(minHeight: 320) `and `.clipped()` from DatePicker, which were workarounds for the insufficient popover height and  are no longer needed

before:
<img width="362" height="475" alt="Screenshot 2026-04-03 at 16 42 20" src="https://github.com/user-attachments/assets/6ea0382e-cca0-43fa-9753-3fa4e0b1a79d" /><img width="365" height="421" alt="Screenshot 2026-04-03 at 14 23 52" src="https://github.com/user-attachments/assets/d705ff5c-cf97-49e5-909e-37436e5564e5" />

after:
<img width="364" height="243" alt="Screenshot 2026-04-07 at 15 14 09" src="https://github.com/user-attachments/assets/1d10c5c9-e74c-49e9-8b74-2e034132e8ab" /><img width="364" height="515" alt="Screenshot 2026-04-07 at 15 14 20" src="https://github.com/user-attachments/assets/0566ced8-ddeb-4dca-abba-b664754913dc" />

### DIFF
--- a/Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
+++ b/Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
@@ -11,11 +11,19 @@ struct CancellableResettableDialogNavigationForm<Title: View, CancelAction: View
     var applyAction: ApplyAction
     let calculateScrollView: Bool
     @State private var sheetContentHeight: CGFloat = 0
+    @State private var hasInitialMeasurement: Bool = false
     @State private var bottomHeight: CGFloat = 0
-    @State private var navigationBarHeight: CGFloat = 50
+    @State private var navigationBarHeight: CGFloat = {
+        if LiquidGlassHelper.usesLiquidGlassUI {
+            return UIDevice.current.userInterfaceIdiom != .phone ? 67 : 74
+        } else {
+            return UIDevice.current.userInterfaceIdiom != .phone ? 50.0 : 56.0
+        }
+    }()
+
     @State private var scrollViewHeight: CGFloat? = nil
     @State private var subScrollViewHeight: CGFloat = 0
-    @State var horizontalSizeClass: UserInterfaceSizeClass = .compact
+    @State var horizontalSizeClass: UserInterfaceSizeClass = UIDevice.current.userInterfaceIdiom != .phone ? .regular : .compact
     
     #if !os(visionOS)
         private let popoverIdealWidth = 393.0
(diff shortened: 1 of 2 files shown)


Sources/FioriSwiftUICore/Components/CancellableResettableForm.swift
```diff
@@ -11,11 +11,19 @@ struct CancellableResettableDialogNavigationForm<Title: View, CancelAction: View
     var applyAction: ApplyAction
     let calculateScrollView: Bool
     @State private var sheetContentHeight: CGFloat = 0
+    @State private var hasInitialMeasurement: Bool = false
     @State private var bottomHeight: CGFloat = 0
-    @State private var navigationBarHeight: CGFloat = 50
+    @State private var navigationBarHeight: CGFloat = {
+        if LiquidGlassHelper.usesLiquidGlassUI {
+            return UIDevice.current.userInterfaceIdiom != .phone ? 67 : 74
```
dyongxu, line 18: is there a way we can calculate the bar height reliably ?
I am not sure if these hardcoded numbers work for all iPhones / iPad models. Also, can they continue to work for future iOS update ?

KevinZK: Currently, hasIdealHeight has already solved the first-frame issue, and the tolerance range for the hardcoded values is very large - even a difference of a few pt would only result in a minor height variation of the sheet. Therefore, the current solution has a very low risk in terms of actual effectiveness.

The old code currently calculates and obtains the height dynamically, and it seems this approach has indeed encountered issues, which is why I resorted to hardcoding.
I can try to add a dynamic correction logic again.

After testing, it was found that adding dynamic calculation still causes problems.

Screenshot 2026-04-08 at 09 59 30

I think the current hardcoding approach is appropriate.

dyongxu (approved):

PR #1709, opened 2026-06-05
fix: 🐛 [2039]Add .deactivate option to FilterFeedbackBar reset

KevinZK:
## Summary
Adds a new `.deactivate` case to `FilterFeedbackBarResetButtonType`, allowing developers to configure the Reset button in a filter chip's popover to deactivate the chip and dismiss the popover, instead of restoring the original value.

This addresses the design requirement that filter chips should have a clear way to be turned off ("inactive" state shows only the chip's name without a value), which previously was not possible for several item types whose values were non-Optional.

### New API: .deactivate reset behavior
Developers can opt in by passing a `FilterFeedbackBarResetButtonConfiguration` with type `.deactivate` to any popover-supporting item:
```swift
.stepper(item: .init(
    name: "Quantity",
    stepperTitle: "Quantity",
    stepRange: 0...100,
    resetButtonConfiguration: FilterFeedbackBarResetButtonConfiguration(with: .deactivate, title: "deactivate")
), showsOnFilterFeedbackBar: true)
```

When the user taps the Reset button, the chip becomes inactive, the popover dismisses, and the chip label reverts to showing only the item's name. Default behavior (`.reset` — restore original value) is unchanged for all existing call sites.

### Behavior matrix
| Reset type | Action | Disabled when |
| ---- | ---- | ---- |
| .reset (default) | Restore working value to original | workingValue == originalValue |
| .clearAll (PickerItem only) | Clear selection | workingValue.isEmpty |
| .deactivate (new) | Clear value + apply + dismiss popover | chip already inactive (!isChecked) |

### Data model changes
- `FilterFeedbackBarResetButtonType`: added `.deactivate` case.
- All 9 popover-supporting item types now have a `resetButtonConfiguration` field and accept it as a constructor parameter (PickerItem already had it; the other 7 types now match this API).
- `mutating func deactivate()` added to all 9 popover types.
- Type widening (breaking-compatible): `StepperItem.value`, `StepperItem.workingValue`, `StepperItem.originalValue` changed from `Double` to `Double?`; `DurationPickerItem.value`, `workingValue`, `originalValue` changed from `Int` to `Int?`. `nil` represents the inactive state. Existing call sites passing literal numbers still compile via Optional promotion.
- `OrderPickerItem` uses an independent `isInactive` flag instead of clearing value, because its value array carries both criterion configuration and per-row selection state — clearing it would lose the criteria themselves.
- `StepperItem.isChecked` and `DurationPickerItem.isChecked` changed from hardcoded `true` to `value != nil`. `OrderPickerItem.isChecked` changed from `true` to `!isInactive`. label for these three returns just the item name when inactive.

### View layer changes
- All 9 popovers' Reset button now dispatches on `resetButtonConfiguration.type` and calls `deactivate() + apply() + onUpdate() + dismiss` when configured for `.deactivate`.
- Reset button disabled state correctly reflects `.deactivate` mode (uses `!isChecked` instead of `isOriginal`).
- Reset button label now reads `resetButtonConfiguration.title` in all 9 popovers (previously only PickerItem honored the configurable title).
- Popover `onAppear` now syncs a placeholder value to `workingValue` when the chip is inactive (DateTime → today; Stepper → stepRange.lowerBound; Slider → lowerBound/upperBound; OrderPicker → resets workingIsInactive flag). This fixes the case where the popover UI shows a placeholder but workingValue is still nil — opening an inactive chip and tapping Apply now activates with the visible placeholder value.
- Stepper and DurationPicker bindings adapted for Optional workingValue (use `?? lowerBound / ?? 0` for display fallback; empty Stepper input maps back to nil).
- `SortFilterCFGItemContainer` (full configuration sheet) updated to handle `.deactivate` in both `handleReset` and the list picker toolbar.

### Bug fixes
- Demo helper `json(item: SortFilterItem.OrderPickerItem)` could crash with "Can't remove last element from an empty collection" when value was empty. Now guards against empty string before `removeLast()`.

## Files Changed
- Sources/FioriSwiftUICore/DataTypes/SortFilter+DataType.swift
- Sources/FioriSwiftUICore/Views/SortFilter/FilterFeedbackBarItemSubview.swift
- Sources/FioriSwiftUICore/Views/SortFilter/SortFilterCFGItemContainer.swift
- Apps/Examples/Examples/FioriSwiftUICore/SortFilter/SortFilterView+Extensions.swift (demo crash fix)
- Apps/Examples/Examples/FioriSwiftUICore/SortFilter/SortFilterExample.swift (added .deactivate examples)

## Backward Compatibility
✅ Source-compatible. All existing call sites continue to compile and behave identically:
- New `resetButtonConfiguration`: init parameters all default to `FilterFeedbackBarResetButtonConfiguration()` (i.e. `.reset` type).
- `StepperItem(value: 5, ...)` and `DurationPickerItem(value: 5, ...)` literal calls still work via Optional promotion.
- `.reset` and `.clearAll` semantics are unchanged.

 ## ⚠️ Breaking Change Notice
This PR contains technically breaking changes that are source-compatible for the vast majority of usage:

1. **`StepperItem.value` / `workingValue` / `originalValue`** changed from `Double` to `Double?`.
2. **`DurationPickerItem.value` / `workingValue` / `originalValue`** changed from `Int` to `Int?`.
3. **`SliderItem` Int convenience init** with `value: nil` no longer coerces to `minimumValue` — it now stays `nil` (chip becomes inactive).

### Migration
Existing code that constructs items with literal values (`value: 5`) continues to work without changes via Swift's Optional promotion. Code that reads `item.value` for arithmetic, comparison, or type conversion needs to unwrap the Optional:

```swift
// Before
let total = stepperItem.value + 10
let displayed = "\(Int(stepperItem.value))"

// After
let total = (stepperItem.value ?? 0) + 10
let displayed = stepperItem.value.map { "\(Int($0))" } ?? ""
```
### Effect

https://github.com/user-attachments/assets/894e21d2-7488-4605-9057-c73ffb1e8fae


dyongxu (dismissed):

dyongxu:
@KevinZK Need to address bot reviews

KevinZK:
> @KevinZK Need to address bot reviews

fixed

dyongxu (approved):

PR #1714, opened 2026-06-17
fix: 🐛 swift 6 error

xiaoyu0722:


Sources/FioriCharts/Charts/AreaMicroChart/AreaMicroChart.swift
```diff
@@ -77,23 +77,3 @@ struct AreaMicroChart: View {
         }
     }
 }
-
```
dyongxu, line 80: Is the removal of preview required here ?

Sources/FioriCharts/TestCases/AreaMicroChartModel+Tests.swift
```diff
@@ -1,103 +0,0 @@
-import SwiftUI
```
dyongxu, line 1: do we have to remove the test?

sourcery/stencils/phase_three/view_modifier_decl.swifttemplate
```diff
@@ -21,7 +21,7 @@ if !variables.isEmpty {
 import SwiftUI
```
dyongxu, line 21: Should sourcery/stencils/ComponentTemplate.swifttemplate be updated as well ?

Sources/FioriCharts/Charts/AreaMicroChart/AreaMicroChart.swift
xiaoyu0722: Seems   `static var previews: some View {`
Syntax error for this static var statement, maybe change to `#Preview` is better

Sources/FioriCharts/TestCases/AreaMicroChartModel+Tests.swift
xiaoyu0722: Mark to @MainActor may be better

sourcery/stencils/phase_three/view_modifier_decl.swifttemplate
xiaoyu0722: Yes
`@MainActor @preconcurrency public protocol ViewModifier`
ViewModifier is main actor isolated, so actually we should not used it in non-isolated environment key.
If insist, we need mark it as main actor isolated too.

xiaoyu0722:
Handled by https://github.com/SAP/cloud-sdk-ios-fiori/pull/1720

PR #1722, opened 2026-07-02
fix: 🐛 [IOSSDKBUG-2144] NoteFormView - When maxTextEditorHeight is set to nil, the full text should always be displayed immediately, without requiring a view re-render.

leoxiao-sap:
Dev Result:
Simulator Screenshot - iPhone 17 Pro Max - 2026-07-02 at 10 48 13

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/NoteFormViewStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/NoteFormViewStyle.fiori.swift
@@ -99,7 +99,6 @@ extension NoteFormViewFioriStyle {
         @Environment(\.isLoading) var isLoading
         @Environment(\.isCustomizedBorder) var isCustomizedBorder
         @Environment(\.noteFormViewCornerRadius) var noteFormViewCornerRadius
-        @State private var textContentHeight: CGFloat = 0
         
         @ViewBuilder
         func makeBody(_ configuration: NoteFormViewConfiguration) -> some View {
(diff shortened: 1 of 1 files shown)


Sources/FioriSwiftUICore/_FioriStyles/NoteFormViewStyle.fiori.swift
```diff
@@ -132,6 +133,16 @@ extension NoteFormViewFioriStyle {
                             self.checkCharCount(configuration, textString: s)
                         }
                         .padding(.bottom, self.isInfoViewNeeded(configuration) ? 0 : 7)
+                        .ifApply(configuration.maxTextEditorHeight == nil) {
+                            $0.modifier(FioriIntrospectModifier<UITextView> { textView in
```
xiaoyu0722: Please avoid to use `FioriIntrospectModifier`, we may remove it later.
Also is there a SwiftUI way to fix this problem, did you try to use LazyVStack for dynamic views instead of VStack in this component?

xiaoyu0722 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/NoteFormViewStyle.fiori.swift
leoxiao-sap: Ok, updated. I tried LazyVStack, but it's unstable.

xiaoyu0722 (dismissed):

hengyi-zhang (approved):

dyongxu (approved):

PR #1751, opened 2026-08-20
fix: 🐛 [IOSSDKBUG-2271] Duration picker does not contain 59 by default

leoxiao-sap:
Before:

https://github.com/user-attachments/assets/cf05f3cc-bb80-4f68-b05f-fa30e52658f4

After:

https://github.com/user-attachments/assets/fbef5302-d085-4996-aaa4-7128bad76aae

### DIFF
--- a/Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
+++ b/Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
@@ -40,13 +40,11 @@ struct DurationPickerViewWrapper: UIViewRepresentable {
                 if let hourIndex = hours.firstIndex(of: hour) {
                     let currentHourRow = pickerView.selectedRow(inComponent: 0)
                     if currentHourRow != hourIndex {
-                        context.coordinator.currentHourIndex = hourIndex
                         pickerView.selectRow(hourIndex, inComponent: 0, animated: false)
                         pickerView.reloadComponent(1)
                     }
                     let expectedMinuteRows = minutesForHour(hourIndex).count
                     if pickerView.numberOfRows(inComponent: 1) != expectedMinuteRows {
-                        context.coordinator.currentHourIndex = hourIndex
                         pickerView.reloadComponent(1)
                     }
                     if let minuteIndex = minutesForHour(hourIndex).firstIndex(of: minute) {
@@ -81,7 +79,6 @@ struct DurationPickerViewWrapper: UIViewRepresentable {
         let minute = self.selection % 60
         
         if let hourIndex = hours.firstIndex(of: hour) {
-            context.coordinator.currentHourIndex = hourIndex
             pickerView.selectRow(hourIndex, inComponent: 0, animated: false)
             if let minuteIndex = minutesForHour(hourIndex).firstIndex(of: minute) {
                 pickerView.selectRow(minuteIndex, inComponent: 1, animated: false)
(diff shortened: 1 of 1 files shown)


Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
```diff
@@ -38,9 +38,22 @@ struct DurationPickerViewWrapper: UIViewRepresentable {
                 let minute = self.selection % 60
 
                 if let hourIndex = hours.firstIndex(of: hour) {
-                    pickerView.selectRow(hourIndex, inComponent: 0, animated: true)
+                    let currentHourRow = pickerView.selectedRow(inComponent: 0)
+                    if currentHourRow != hourIndex {
+                        context.coordinator.currentHourIndex = hourIndex
+                        pickerView.selectRow(hourIndex, inComponent: 0, animated: false)
+                        pickerView.reloadComponent(1)
```
xiaoyu0722: `pickerView.selectRow` is already called, should we still call `reloadComponent(0)`?

Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
```diff
@@ -94,6 +108,7 @@ struct DurationPickerViewWrapper: UIViewRepresentable {
     class Coordinator: NSObject, UIPickerViewDelegate, UIPickerViewDataSource {
         var parent: DurationPickerViewWrapper
         let pickerFont = UIFont.preferredFioriFont(fixedSize: 22)
+        var currentHourIndex: Int = 0
```
xiaoyu0722: Why we need this? The hour index should be bind with picker view index.
Can't we get it from parent directly? We will always need maintain this property sync with picker index, not sure if it's a good idea.

xiaoyu0722 (changes requested):
I did some investigation for this bug, seems this only happened for the initialization.
So I prefer to check if the number of rows is same with minutes count of selected hour.
So you can try below way.
```swift
func updateUIView(_ uiView: UIViewType, context: Context) {
	for subView in uiView.subviews where subView is UIPickerView {
		if let pickerView = subView as? UIPickerView {
			let hour = self.selection / 60
			let minute = self.selection % 60

			if let hourIndex = hours.firstIndex(of: hour) {
                                 // New added logic.
				if pickerView.numberOfRows(inComponent: 1) != minutesForHour(hourIndex).count {
					pickerView.reloadComponent(1)
				}
				pickerView.selectRow(hourIndex, inComponent: 0, animated: true)
				if let minuteIndex = minutesForHour(hourIndex).firstIndex(of: minute) {
					pickerView.selectRow(minuteIndex, inComponent: 1, animated: true)
				}
			}
			break
		}
	}
}
```

Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
leoxiao-sap: Tested and updated

Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
leoxiao-sap: Removed

Sources/FioriSwiftUICore/DurationPicker/DurationPickerViewWrapper.swift
```diff
@@ -121,27 +131,33 @@ struct DurationPickerViewWrapper: UIViewRepresentable {
...
+                return minutes.count > row ? minutes[row].description : ""
             default:
                 return ""
             }
         }
         
         func pickerView(_ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
             if component == 0 {
+                let hourIndex = pickerView.selectedRow(inComponent: 0)
+                var minuteRow = pickerView.selectedRow(inComponent: 1)
```
xiaoyu0722, line 144: Since, selected row in component 0 is `hourIndex`, why this not `minuteIndex`

xiaoyu0722 (approved):

KevinZK (approved):

PR #1764, opened 2026-09-02
fix: 🐛 [2278]Fixed FilterFormView's item padding

KevinZK:
Spacing of long text in filter button gets pinched.
Modify the padding values by referring to the design draft.
Simulator Screenshot - iPhone 17 - 2026-09-02 at 16 29 53

### DIFF
--- a/Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
+++ b/Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
@@ -1554,7 +1554,7 @@ public enum ListPickerSearchBarDisplayMode {
 }
 
 struct ListPickerSearchBarDisplayModeKey: EnvironmentKey {
-    static let defaultValue: ListPickerSearchBarDisplayMode = .automatic
+    static let defaultValue: ListPickerSearchBarDisplayMode = .always
 }
 
 public extension EnvironmentValues {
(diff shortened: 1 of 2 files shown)


Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
```diff
@@ -1554,7 +1554,7 @@ public enum ListPickerSearchBarDisplayMode {
 }
 
 struct ListPickerSearchBarDisplayModeKey: EnvironmentKey {
-    static let defaultValue: ListPickerSearchBarDisplayMode = .automatic
+    static let defaultValue: ListPickerSearchBarDisplayMode = .always
```
xiaoyu0722, line 1557: What's wrong with automatic?
If this is not supported, should we remove this api? If supported, then the api doc should also be updated.

xiaoyu0722 (changes requested):

Sources/FioriSwiftUICore/_FioriStyles/ListPickerDestinationStyle.fiori.swift
KevinZK: 1. Under automatic mode, the original ListPickerSearchBarDisplayMode was set to .always.
2. When automatic is the default setting, the sheet within the sortFilter component experiences stuttering on the first frame.
Workaround.

xiaoyu0722 (approved):

PR #1767, opened 2026-09-03
fix: 🐛 [IOSSDKBUG-2296] Add rounded corners for the swipe-revealed view of the ObjectItem

leoxiao-sap:
Dev results:
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-02 at 15 55 29
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-02 at 15 54 41
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-02 at 15 55 10
Simulator Screenshot - iPhone 17 Pro Max - 2026-09-02 at 16 12 40

### DIFF
--- a/Apps/Examples/Examples/FioriSwiftUICore/ObjectItem/ObjectItemListView.swift
+++ b/Apps/Examples/Examples/FioriSwiftUICore/ObjectItem/ObjectItemListView.swift
@@ -12,6 +12,11 @@ struct ObjectItemListView<T: ListDataProtocol>: View {
     @State var cellTapped = false
     @State var singleSelection: Int?
     
+    /// Corner radius used for the swipe-revealed rounded corner.
+    private let rowCornerRadius: CGFloat = 16
+    
+    private let cellHorizontalInset: CGFloat = 32
+    
     init(title: String, listDataType: T.Type, changeLeftMargin: Bool = true, showEditButton: Bool = true) {
         self.title = title
         self.listDataType = listDataType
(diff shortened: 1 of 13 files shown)


Apps/Examples/Examples/FioriSwiftUICore/ObjectItem/ObjectItemListView.swift
```diff
@@ -27,30 +35,44 @@ struct ObjectItemListView<T: ListDataProtocol>: View {
         }
     }
     
+    @ViewBuilder
+    private func swipeRoundedTrailing(@ViewBuilder content: () -> some View) -> some View {
+        content()
+            .padding(.horizontal, self.cellHorizontalInset)
+            .padding(.vertical, self.cellVerticalInset)
+            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .leading)
```
xiaoyu0722, line 43: Does the frame setting needed?
Also usually, using modifier or update styles directly, but not pass a view.
Like
```swift
func itemContent(listData: ListData, _ indexPath: IndexPath) {
    listData.cellForRow(indexPath)
        .padding...
        .background...
        .clipShape...
}
```

xiaoyu0722 (changes requested):

Apps/Examples/Examples/FioriSwiftUICore/ObjectItem/ObjectItemListView.swift
leoxiao-sap: Yes, we need to set the frame, otherwise, when swiping left on a row, the background color will separate from the content size when the content does not fill the row(eg: Compact View Examples). Using the usual way will cause the code to be repeated twice, because listData.cellForRow appears in an if-else pattern.
image

xiaoyu0722 (approved):

